PR #7, opened 2022-02-23
feat: Replace the serial number with endpointReference address UUID

### LINKED ISSUE #2: Device Discovery should use EndpointReference for uniqueness
**Observation**
We have discovered that our `Tapo C200` cameras all return the same serial number when queried via `GetDeviceInformation`.
We have also learned that the `happytime-multi-onvif-server` sets all camera serial numbers to `123456` by default.

**Problem**
Only 1 device of each type `Tapo C200` and `happytime` can be added to EdgeX due to name conflicts.

**Proposed Solution**

Use the `Endpoint Reference` as specified in the ONVIF spec. This is available during WS-Discovery:

> 
> **7.3.1 Endpoint reference**
> A device or an endpoint that takes the client role should use a URN:UUID [RFC4122] as the address property
> of its endpoint reference.
> 
> The device or an endpoint that takes the client role shall use a **stable, globally unique identifier that is constant
> across network interfaces** as part of its endpoint reference property. The combination of an wsadis:Address
> and wsadis:ReferenceProperties provide a stable and globally-unique identifier.

[Source](https://www.onvif.org/specs/core/ONVIF-Core-Specification.pdf)

We have indeed confirmed that even when the device IP address changes, the `EndpointReference` stays the same, and is unique to all of our cameras.

weichou1229:
Close #2

Signed-off-by: bruce <weichou1229@gmail.com>

## PR Checklist

- [ ] Tests for the changes have been added (for bug fixes / features) **not impact the teat**
- [x] Docs have been added / updated (for bug fixes / features)

The serial number is the same for the different TAPO C200 cameras, the device name would be duplicated.

## What is the new behavior?
Replace the serial number with EndpointReference Address UUID as the part of the device name

- [ ] Yes
- [x] No

## New Imports

## Specific Instructions

## Other information

### DIFF
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -299,6 +299,10 @@ func (d *Driver) Discover() {
 	onvifDevices := wsdiscovery.GetAvailableDevicesAtSpecificEthernetInterface(d.config.DiscoveryEthernetInterface)
 	var discoveredDevices []sdkModel.DiscoveredDevice
 	for _, onvifDevice := range onvifDevices {
+		if onvifDevice.GetDeviceParams().EndpointRefAddress == "" {
+			d.lc.Warnf("The EndpointRefAddress is empty from the Onvif camera, unable to add the camera %s", onvifDevice.GetDeviceParams().Xaddr)
+			continue
+		}
 		address, port := addressAndPort(onvifDevice.GetDeviceParams().Xaddr)
 		dev := models.Device{
 			// Using Xaddr as the temporary name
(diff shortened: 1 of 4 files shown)


internal/driver/driver.go
```diff
@@ -325,7 +325,7 @@ func (d *Driver) Discover() {
 		dev.Protocols[OnvifProtocol][SerialNumber] = devInfo.SerialNumber
 		dev.Protocols[OnvifProtocol][HardwareId] = devInfo.HardwareId
 		discovered := sdkModel.DiscoveredDevice{
-			Name:        fmt.Sprintf("%s-%s-%s", devInfo.Manufacturer, devInfo.Model, devInfo.SerialNumber),
+			Name:        fmt.Sprintf("%s-%s-%s", devInfo.Manufacturer, devInfo.Model, onvifDevice.GetDeviceParams().UUID),
```
ajcasagrande: As I mentioned [in the review of the onvif library PR](https://github.com/IOTechSystems/onvif/pull/8#discussion_r813424628), we may want to handle the empty UUID case in this code.

weichou1229: Modified. 
@ajcasagrande  I also make changes to the PR https://github.com/IOTechSystems/onvif/pull/8, could you help review that?

ajcasagrande (approved):
LGTM

cloudxxx8:
please upgrade go.mod to use the latest onvif library

weichou1229:
Updated the onvif library.

cloudxxx8 (approved):
LGTM

PR #10, opened 2022-03-09
feat: Auto-Discovery should work for unauthorized devices

### LINKED ISSUE #3: Discovery fails for unauthorized devices
**Observation**
When discovering devices, if the device is password-protected, and the credentials do not match the default credentials, the device will not be added to EdgeX. This is because the `GetDeviceInformation` requires authentication.

**Problem**
Without these devices being added to EdgeX there is no way to assign credentials to them. Thus, currently your solution requires all auto-discovered cameras to share the same username, password, and authentication type (which is bad practice).

**Proposed Solution**
- If a device fails the `GetDeviceInformation`, but it is still a valid ONVIF device, it should be added to EdgeX, so that the credentials can be added after the fact.
- Fix #2 and you will be able to add devices with unique names even without authentication.

weichou1229:
Close #3

Signed-off-by: bruce <weichou1229@gmail.com>

## PR Checklist

- [ ] Tests for the changes have been added (for bug fixes / features) **not impact the test**
- [ ] Docs have been added / updated (for bug fixes / features) **not impact the doc**

If a device fails the GetDeviceInformation, the device can't be added to the EdgeX.

## What is the new behavior?
If a device fails GetDeviceInformation, it should be added to EdgeX, so that the credentials can be added after the fact.

- [ ] Yes
- [x] No

## New Imports

## Specific Instructions

## Other information

### DIFF
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -318,23 +318,32 @@ func (d *Driver) Discover() {
 		}
 
 		devInfo, edgexErr := d.getDeviceInformation(dev)
+		endpointRef := onvifDevice.GetDeviceParams().EndpointRefAddress
+		var discovered sdkModel.DiscoveredDevice
 		if edgexErr != nil {
-			d.lc.Errorf("failed to get the device information for the camera %s, %v", dev.Name, edgexErr)
-			return
+			d.lc.Warnf("failed to get the device information for the camera %s, %v", endpointRef, edgexErr)
+			dev.Protocols[OnvifProtocol][SecretPath] = endpointRef
+			discovered = sdkModel.DiscoveredDevice{
+				Name:        endpointRef,
+				Protocols:   dev.Protocols,
+				Description: "Auto discovered Onvif camera",
+				Labels:      []string{"auto-discovery"},
+			}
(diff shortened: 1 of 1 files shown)


internal/driver/driver.go
```diff
@@ -318,23 +318,31 @@ func (d *Driver) Discover() {
...
-			d.lc.Errorf("failed to get the device information for the camera %s, %v", dev.Name, edgexErr)
-			return
+			d.lc.Warnf("failed to get the device information for the camera %s, %v", endpointRef, edgexErr)
+			dev.Protocols[OnvifProtocol][SecretPath] = endpointRef
+			discovered = sdkModel.DiscoveredDevice{
+				Name:        endpointRef,
+				Protocols:   dev.Protocols,
+				Description: "Auto discovered Onvif camera",
+				Labels:      []string{"auto-discovery"},
+			}
```
ajcasagrande, line 331:
```suggestion
			}
			d.lc.Debugf("Discovered unknown camera from the address '%s'", onvifDevice.GetDeviceParams().Xaddr)
```

ajcasagrande (commented):
Looks good. One suggestion.

internal/driver/driver.go
weichou1229: Added the debug log.

ajcasagrande (approved):
LGTM

PR #21, opened 2022-03-26
updated auto-discovery-docker documentation

vyshali-chitikeshi:
Below things needs to update in auto-discover-docker documentation:
1. Device onvif camera image path should change to edgexfoundry/
2. Command needs to have -cp=consul.http://localhost:8500 for credentials to sync with device service after user provides camera credentials. Without this, onvif device service not able to retrieve camera credentials from consul and api's are not working.


ajcasagrande (approved):
LGTM

PR #43, opened 2022-04-15
fix: override server bind address to 0.0.0.0

ajcasagrande:
- this will allow it to still be called via localhost and not just the
  external IP address

Signed-off-by: Anthony Casagrande <anthony.j.casagrande@intel.com>


cloudxxx8 (approved):
LGTM

PR #48, opened 2022-04-26
Feature: Device control plane events

ajcasagrande:
## PR Checklist

- [ ] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)

Only the device service knows when a camera is added, updated, or deleted.

## Issue Number:
N/A

## What is the new behavior?

When a device is added, updated, or deleted from the device service, an event will be added to the EdgeX message bus letting other services be notified of the change.

- [ ] Yes
- [X] No

## New Imports
No
## Specific Instructions
No

## Other information
This code is in lieu of official control plane event support which will be coming in a later release of EdgeX.


vchitike:
Validation looks good. Verified that device addition, update and delete events are emitting as expected.

cmd/res/profiles/control-plane.profile.yaml
```diff
@@ -0,0 +1,29 @@
+name: "device-onvif-camera"
```
presatish, line 1: Will it be better to name this with `control-plane` something? The two existing profiles in the profiles dir have pretty much same kind of naming.

internal/driver/driver.go
```diff
@@ -77,14 +88,15 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModel.As
...
-	handler := NewRestNotificationHandler(sdk.RunningService(), lc, asyncCh)
+	handler := NewRestNotificationHandler(deviceService, lc, asyncCh)
 	edgexErr := handler.AddRoute()
 	if edgexErr != nil {
 		return errors.NewCommonEdgeXWrapper(edgexErr)
 	}
 
 	d.lc.Info("Driver initialized.")
 	return nil
+
```
presatish: extra line needs to be removed.

presatish (approved):
Have provided some suggestions. Overall LGTM.

cmd/res/profiles/control-plane.profile.yaml
ajcasagrande: The reason we named it as such is because there are scripts and services that may need to filter out this device when querying for all devices (it gets returned as if it is a camera). If it was named as `control-plane` something, the name would need to be known by all of those scripts and services, and it would need to be not changed as to avoid breaking anything.

Because of this, we felt that naming it the same as the device service would allow anyone that knows the name of the service to know the name of the control-plane device.

internal/driver/driver.go
ajcasagrande: removed

cmd/res/profiles/control-plane.profile.yaml
```diff
@@ -0,0 +1,29 @@
...
+manufacturer:  "Intel"
+model: "Control Plane Device for device-onvif-camera device service"
+labels:
+  - "onvif"
+  - "control-plane"
+description: "EdgeX device profile for device-onvif-camera Control Plane Events."
+
+deviceResources:
+  - name: "CameraAdded"
+    isHidden: true
```
lenny-goodell: I like to put a comment on this line for each resource stating that the resource is Async only, co no commands.

doc/auto-discovery-docker.md
```diff
@@ -29,11 +29,11 @@ services:
     network_mode: "host"
     environment:
       SERVICE_HOST: 192.168.56.101
+      SERVICE_SERVERBINDADDR: 0.0.0.0
```
lenny-goodell: I thought we are getting rid of this and documenting separately how to configure to the ONVIF rest callback notifications. i.e we don't want to open this up by default, only when the user wants ONVIF rest callback and knows the risks of opening it up like this.

So for now this should be removed and service `SERVICE_HOST: edgex-device-onvif-camera` and  remove `network_mode: "host"`

doc/control-plane-events.md
```diff
@@ -0,0 +1,175 @@
...
+                  "FirmwareVersion":"2.4a",
+                  "HardwareId":"1.0",
+                  "Manufacturer":"Intel",
+                  "Model":"SimCamera",
+                  "Port":"10000",
+                  "SecretPath":"credentials002",
+                  "SerialNumber":"c45a57b3"
+               }
+            },
+            "ServiceName":"device-onvif-camera"
```
lenny-goodell: Is all this data needed? or can we just have the new device name as a simple string value? 
```
         "valueType":"String",
         "value":"Intel-SimCamera-c45a57b3-6fcb-4c51-83de-54495e5454a6"
```
If not, what are we using all this info for. If the receiver need more data they can query Core Metadata for the details.
I assume for Skyline trail we just device name to use in Core Command that is needed.

internal/driver/driver.go
```diff
@@ -65,6 +72,10 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModel.As
 	deviceService := sdk.RunningService()
 
 	for _, device := range deviceService.Devices() {
+		if device.Name == d.serviceName {
```
lenny-goodell, line 76: Add "why" comment here

internal/driver/driver.go
```diff
@@ -217,23 +228,75 @@ func (d *Driver) Stop(force bool) error {
...
+		if err != nil {
+			d.lc.Errorf("issue getting device %s: %v", eventType, deviceName, err)
+			return
+		}
+
+		cv, err = sdkModel.NewCommandValue(eventType, common.ValueTypeObject, dev)
+		if err != nil {
+			d.lc.Errorf("issue creating control-plane event %s for device %s: %v", eventType, deviceName, err)
+			return
+		}
```
lenny-goodell: IMO, should just be the same as cameraDeleted

internal/driver/driver.go
```diff
@@ -217,23 +228,75 @@ func (d *Driver) Stop(force bool) error {
...
+	d.asynchCh <- asyncValues
+}
+
 // AddDevice is a callback function that is invoked
 // when a new Device associated with this Device Service is added
 func (d *Driver) AddDevice(deviceName string, protocols map[string]models.ProtocolProperties, adminState models.AdminState) error {
-	err := d.createOnvifClient(deviceName)
-	if err != nil {
-		return errors.NewCommonEdgeXWrapper(err)
+	// only execute if this was not called for the control-plane device
```
lenny-goodell, line 252: Nice `why` comment. :-)

lenny-goodell (changes requested):
Nice job. Just a few comments...

doc/auto-discovery-docker.md
ajcasagrande: Yes, I was a little confused as to why this code was showing as being new/different, as it was originally changed in main branch. I think maybe a rebase caused to to be seen as a different commit. We can remove it.

doc/control-plane-events.md
ajcasagrande: Originally that is what we had, but I decided to send the whole device in case they needed it. Thinking back over it, you are right, they can just query the metadata service for it. My initial hesitation from that was probably linked to an issue we had in the past with the LLRP service, but this one will not suffer from that due to the events being sent _after_ the device service cache is updated.

cmd/res/profiles/control-plane.profile.yaml
ajcasagrande: @brianointel here is an example:
https://github.com/marcpfuller/device-rfid-llrp-go/blob/7b55924c09c2d909d14f4730cd7f1b6443ec55e3/cmd/res/profiles/llrp.device.profile.yaml#L64

lenny-goodell (approved):
LGTM

ajcasagrande (commented):
LGTM

weichou1229:
LGTM.
Since the support notification can send `rest` or `email` when a device has been added, updated, or deleted,

https://github.com/edgexfoundry/edgex-go/blob/ff693608c36af4ab72a5c512cfc781335cb747c6/internal/core/metadata/application/notify.go#L60

https://github.com/edgexfoundry/edgex-go/blob/ff693608c36af4ab72a5c512cfc781335cb747c6/internal/core/metadata/application/notify.go#L110

https://github.com/edgexfoundry/edgex-go/blob/ff693608c36af4ab72a5c512cfc781335cb747c6/internal/core/metadata/application/notify.go#L130

 is it possible to enhance the support notification to emit events onto the message bus? or this is a different feature to notification service?

lenny-goodell:
> Since the support notification can send `rest` or `email` when a device has been added, updated, or deleted,

@weichou1229 , this is temporary until the [Control Plane Event ADR](https://github.com/edgexfoundry/edgex-docs/pull/724) is completed, Approved and Implemented. Your questions above are appropriate to the ADR decision. Please review it and add your comments their.

weichou1229 (approved):
Got it. Thanks for your info about the CPE usage.

PR #65, opened 2022-05-18
Discovery probe

### LINKED ISSUE #39: Improve auto discovery mechanism
Improve the network scan for discover devices
https://github.com/edgexfoundry-holding/device-onvif-camera/pull/13#discussion_r836663483

ajcasagrande:
## PR Checklist

- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)

Multicast only ws-discovery

## Issue Number:
Closes #39 

## What is the new behavior?

Support for dual discovery modes: multicast and netscan. Netscan works by scanning a range of IP addresses with a UDP
unicast probe on the ws-discovery port 3702. Netscan allows the service to discover devices using docker without the need for
network_mode: host.

- [ ] Yes
- [X] No

## New Imports

- self-owned netscan package
- pkg/errors for wrapping errors

## Specific Instructions

Run the following command to determine what to put for `DRIVER_DISCOVERYSUBNETS`:

```bash
ip -4 -o route list scope link | sed -En "s/ dev ($(find /sys/class/net -mindepth 1 -maxdepth 2 -not -lname '*devices/virtual*' -execdir grep -q 'up' "{}/operstate" \; -printf '%f\n' | paste -sd\| -)).+//p" | grep -v "169.254.0.0/16" | sort -u | paste -sd, -
```

## Other information


ajcasagrande:
@weichou1229 @cloudxxx8 @lenny-intel can the DCO be overriden to pass? I believe it is failing incorrectly, as it is unable to properly match @ChristianDarr-personal 's commit, but the commit is signed off by him.

lenny-goodell:
> can the DCO be overriden to pass?
@ajcasagrande , yup. Done

lenny-goodell:
recheck

lenny-goodell:
recheck

lenny-goodell:
recheck

lenny-goodell:
@ajcasagrande , please squash your commits and sign-off the new commit message. This will fix the DCO failure.

ajcasagrande:
@lenny-intel @weichou1229 @cloudxxx8 can someone code review this, please? I don't have permisions to assign reviewers.

Btw, the reason I have organized the code as such (adding a sub go-module for netscan vs just a go package) is that the code is based on the LLRP discovery code I wrote, and the intention would be for both of them to share it (whether that means it gets its own repo, or what). I also have proof of concept code that migrates the LLRP device service to implement `netscan.ProtocolSpecificDiscovery`.

lenny-goodell:
> Btw, the reason I have organized the code as such (adding a sub go-module for netscan vs just a go package) is that the code is based on the LLRP discovery code I wrote, and the intention would be for both of them to share it (whether that means it gets its own repo, or what). I

@ajcasagrande , @cloudxxx8 , @weichou1229 , should this `netscan` code be integrated in to the Device SDK, rather than embedded in this repo?

It will need to be in this repo until we release this service for Kamakura. After that it should be integrated in the Device SDK.

cmd/res/configuration.toml
```diff
@@ -108,8 +108,26 @@ UseMessageBus = true
...
+# Maximum simultaneous network probes
+ProbeAsyncLimit = "4000"
+
+# Maximum amount of milliseconds to wait for each IP probe before timing out.
+# This will also be the minimum time the discovery process can take.
+ProbeTimeoutMillis = "2000"
+
+# Maximum amount of seconds the discovery process is allowed to run before it will be cancelled.
+# It is especially important to have this configured in the case of larger subnets such as /16 and /8
+MaxDiscoverDurationSeconds = "300"
```
lenny-goodell, line 133: The new custom config would be better here so the numbers don't have to be strings requiring code to parse them.

doc/auto-discovery-docker.md
```diff
@@ -1,60 +1,49 @@
...
 
-## None Security Mode
+## Non-Security Mode
 
 ### Prepare edgex-compose/compose-builder
 
 #### 1. Download the [edgex-compose](https://github.com/edgexfoundry/edgex-compose) and setup it according to the [docker-compose setup guide](./docker-compose/README.md)
 
-#### 2. Replace the `add-device-onvif-camera.yml` with the following content:
+#### 2. Update  the `add-device-onvif-camera.yml` file with the following content:
```
lenny-goodell, line 15: Please rework this to not duplicate the ymal that is now part of compose builder. Should only have snippet of what additional env overrides should be added. See this issue: https://github.com/edgexfoundry/device-onvif-camera/issues/69

doc/auto-discovery-docker.md
```diff
@@ -89,51 +78,38 @@ Then user can follow [the doc to add Provision Watcher](./auto-discovery.md).
 
 #### 2. Replace the `add-device-onvif-camera.yml` with the following content:
```
lenny-goodell, line 79: Please rework this to not duplicate the ymal that is now part of compose builder. Should only have snippet of what additional env overrides should be added. See this issue: https://github.com/edgexfoundry/device-onvif-camera/issues/69

doc/auto-discovery-docker.md
```diff
@@ -89,51 +78,38 @@ Then user can follow [the doc to add Provision Watcher](./auto-discovery.md).
...
-- Add `EDGEX_SECURITY_SECRET_STORE` env with "true" value
-- Enable auto-discovery by `DEVICE_DISCOVERY_ENABLED` with "true" value
-- Use `DRIVER_DISCOVERYETHERNETINTERFACE` to specify the ethernet interface for discovering
-- Use `DRIVER_DEFAULTSECRETPATH` to specify the default secret path
-- Use `SECRETSTORE_HOST` to specify the Vault's host
-- Use `STAGEGATE_BOOTSTRAPPER_HOST`, `STAGEGATE_READY_TORUNPORT`, `STAGEGATE_WAITFOR_TIMEOUT` to specify the bootstrapper settings for waiting the security set up
-- Use `WRITABLE_LOGLEVEL` to specify the log level for debugging
-- Add command to override the CMD because we don't use the configuration provider from Consul
-
 #### 3. Export the Security Bootstrapper
```
lenny-goodell, line 113: Why is this being done?

internal/driver/driver.go
```diff
@@ -70,22 +83,26 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModel.As
 	d.config = camConfig
 
 	deviceService := sdk.RunningService()
+	d.svc = &DeviceSDKService{
+		DeviceService: deviceService,
+		lc:            lc,
+	}
 
-	for _, device := range deviceService.Devices() {
+	for _, dev := range deviceService.Devices() {
```
lenny-goodell: why shorten name when `device` is a better name for readability?

internal/driver/driver.go
```diff
@@ -325,78 +353,86 @@ func (d *Driver) getCredentials(secretPath string) (config.Credentials, errors.E
...
+// multicast enable/disable via config option
+func (d *Driver) discoverMulticast(discovered []sdkModel.DiscoveredDevice) []sdkModel.DiscoveredDevice {
+	t0 := time.Now()
 	onvifDevices := wsdiscovery.GetAvailableDevicesAtSpecificEthernetInterface(d.config.DiscoveryEthernetInterface)
-	var discoveredDevices []sdkModel.DiscoveredDevice
+	d.lc.Infof("Discovered %d device(s) in %v via multicast.", len(onvifDevices), time.Since(t0))
 	for _, onvifDevice := range onvifDevices {
-		if onvifDevice.GetDeviceParams().EndpointRefAddress == "" {
-			d.lc.Warnf("The EndpointRefAddress is empty from the Onvif camera, unable to add the camera %s", onvifDevice.GetDeviceParams().Xaddr)
+		dev, err := d.createDiscoveredDevice(onvifDevice)
```
lenny-goodell:
```suggestion
		device, err := d.createDiscoveredDevice(onvifDevice)
```

internal/driver/driver.go
```diff
@@ -325,78 +353,86 @@ func (d *Driver) getCredentials(secretPath string) (config.Credentials, errors.E
...
-		}
-		discoveredDevices = append(discoveredDevices, discovered)
+	return discovered
+}
+
+// netscan enable/disable via config option
+func (d *Driver) discoverNetscan(ctx context.Context, discovered []sdkModel.DiscoveredDevice) []sdkModel.DiscoveredDevice {
+
+	if len(strings.TrimSpace(d.config.DiscoverySubnets)) == 0 {
+		d.lc.Debug("discovery not performed, DiscoverySubnets are empty.")
```
lenny-goodell:
```suggestion
		d.lc.Warn("discovery not performed, DiscoverySubnets are empty.")
```

internal/driver/driver.go
```diff
@@ -325,78 +353,86 @@ func (d *Driver) getCredentials(secretPath string) (config.Credentials, errors.E
...
+	result := netscan.AutoDiscover(ctx, NewOnvifProtocolDiscovery(d), params)
+	if ctx.Err() != nil {
+		d.lc.Warn("Discover process has been cancelled!", "ctxErr", ctx.Err())
+	}
+
+	d.lc.Debugf("NetScan result: %+v", result)
+	d.lc.Infof("Discovered %d device(s) in %v via netscan.", len(result), time.Since(t0))
+
+	for _, res := range result {
+		dev, ok := res.Info.(sdkModel.DiscoveredDevice)
```
lenny-goodell:
```suggestion
		device, ok := res.Info.(sdkModel.DiscoveredDevice)
```

internal/driver/driver.go
```diff
@@ -426,21 +462,22 @@ func (d *Driver) getDeviceInformation(dev models.Device) (devInfo *device.GetDev
 }
 
 // newOnvifClient creates a temporary client for auto-discovery
-func (d *Driver) newTemporaryOnvifClient(device models.Device) (*OnvifClient, errors.EdgeX) {
-	cameraInfo, edgexErr := CreateCameraInfo(device.Protocols)
+func (d *Driver) newTemporaryOnvifClient(dev models.Device) (*OnvifClient, errors.EdgeX) {
```
lenny-goodell:
```suggestion
func (d *Driver) newTemporaryOnvifClient(device models.Device) (*OnvifClient, errors.EdgeX) {
```

internal/driver/driver.go
```diff
@@ -426,21 +462,22 @@ func (d *Driver) getDeviceInformation(dev models.Device) (devInfo *device.GetDev
...
+		// since this is just a temporary client, we do not want to wait for credentials to be available
+		credential, edgexErr = d.tryGetCredentials(cameraInfo.SecretPath)
 		if edgexErr != nil {
-			return nil, errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf("failed to get credentials for camera %s", device.Name), edgexErr)
+			return nil, errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf("failed to get credentials for camera %s", dev.Name), edgexErr)
 		}
 	}
 
-	dev, err := onvif.NewDevice(onvif.DeviceParams{
+	onvifDevice, err := onvif.NewDevice(onvif.DeviceParams{
```
lenny-goodell, line 468: I like the explicit name! ;-)

internal/driver/service_test.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+//func (s *MockSDKService) clearDevices() {
+//	s.devices = make(map[string]models.Device)
+//	s.resetAddedCount()
+//}
+
+//func (s *MockSDKService) resetAddedCount() {
+//	atomic.StoreUint32(&s.added, 0)
+//}
```
lenny-goodell: Remove commented out code

internal/driver/service.go
```diff
@@ -0,0 +1,37 @@
...
+
+package driver
+
+import (
+	"github.com/edgexfoundry/device-sdk-go/v2/pkg/service"
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/clients/logger"
+	contract "github.com/edgexfoundry/go-mod-core-contracts/v2/models"
+)
+
+// ServiceWrapper wraps an EdgeX SDK service so it can be easily mocked in tests.
```
lenny-goodell: Please submit an issue on the Device SDK to use interface so you don't have to create this wrapper on every service you want to be able to properly unit test

internal/driver/service_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/pkg/errors"
+)
+
+type MockSDKService struct {
+	devices map[string]models.Device
+	added   uint32
+	Config  map[string]string
+}
+
+func NewMockSdkService() *MockSDKService {
```
lenny-goodell: Preference is to use mockery (https://github.com/vektra/mockery) to generate mocks from interface definitions.
Example: https://github.com/edgexfoundry/app-functions-sdk-go/blob/main/pkg/interfaces/mocks/ApplicationService.go

pkg/netscan/types.go
```diff
@@ -0,0 +1,90 @@
...
+type DiscoveredDevice struct {
+	Name string
+	Info interface{}
+}
+
+// Logger is a generic logging interface in order to not lock this code
+// into a specific logging framework. It is directly compatible with, but not limited to:
+// - github.com/edgexfoundry/go-mod-core-contracts/v2/clients/logger LoggingClient
+// - github.com/sirupsen/logrus Logger
+type Logger interface {
```
lenny-goodell: Why do we need this and not use our logger client interface here: https://github.com/edgexfoundry/go-mod-core-contracts/blob/main/clients/logger/logger.go#L35

pkg/netscan/util_test.go
```diff
@@ -0,0 +1,142 @@
...
+	close(ipCh)
+	wg.Wait()
+
+	return result
+}
+
+// TestIpGenerator calls the ip generator and validates that the first ip, last ip, and size
+// match the expected values.
+func TestIpGenerator(t *testing.T) {
+	tests := []inetTest{
```
lenny-goodell: Need an error test case. Looks like test code looks for it, but missing actual tests case.

pkg/netscan/util_test.go
```diff
@@ -0,0 +1,142 @@
...
+				if result.size != input.size {
+					t.Errorf("expected %d ips, but got %d", input.size, result.size)
+				}
+				if result.first != input.first {
+					t.Errorf("expected first ip in range to be %s, but got %s", input.first, result.first)
+				}
+				if result.last != input.last {
+					t.Errorf("expected last ip in range to be %s, but got %s", input.last, result.last)
+				}
+			}
```
lenny-goodell:
```suggestion
			if input.err {
			   require.True(t,result.err, "expected an error, but no error was returned")
			} else {
			   require.False(t,result.err, "got unexpected error")
                       }
                       
                       assert.Equal(t, input.size, result.size)
                       assert.Equal(t, input.first, result.first)
                       assert.Equal(t, input.last, result.last)
```
require terminates the test if condition fails
If assert fails it displays the expected and actual values, so no need for additional complex message.

pkg/netscan/util_test.go
```diff
@@ -0,0 +1,142 @@
...
+func TestIpGeneratorSubnetSizes(t *testing.T) {
+	// stop at 10 because the time taken gets exponentially longer
+	for i := 32; i >= 10; i-- {
+		i := i
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			t.Parallel()
+			result := ipGeneratorTest(inetTest{size: uint32(i), inet: fmt.Sprintf("192.168.1.1/%d", i)})
+			if result.size != computeNetSz(i) {
+				t.Errorf("expected %d ips, but got %d", computeNetSz(i), result.size)
+			}
```
lenny-goodell:
```suggestion
			assert.Equal(t, computeNetSz(i), result.size)
```

lenny-goodell (changes requested):
First pass review. Sill need to review:
- internal/driver/onvifdiscovery.go
- pkg/netscan/discover.go

internal/driver/onvifdiscovery.go
```diff
@@ -0,0 +1,235 @@
...
+	return ports
+}
+
+// OnConnectionDialed handles the protocol specific verification if there is actually
+// a valid device or devices at the other end of the connection.
+func (proto *OnvifProtocolDiscovery) OnConnectionDialed(host string, port string, conn net.Conn, params netscan.Params) ([]netscan.ProbeResult, error) {
+	// attempt a basic direct probe approach using the open connection
+	devices, err := executeRawProbe(conn, params)
+	if err != nil {
+		params.Logger.Debugf(err.Error())
```
lenny-goodell: Comment why using Debug and not Error or Warn
```suggestion
		params.Logger.Debug(err.Error())
```

internal/driver/onvifdiscovery.go
```diff
@@ -0,0 +1,235 @@
...
+				AuthMode:           d.config.DefaultAuthMode,
+				SecretPath:         d.config.DefaultSecretPath,
+				EndpointRefAddress: endpointRefAddr,
+			},
+		},
+	}
+
+	devInfo, edgexErr := d.getDeviceInformation(dev)
+	if edgexErr != nil {
+		// try again using the device name as the SecretPath
```
lenny-goodell: Comment why need to do this

internal/driver/onvifdiscovery.go
```diff
@@ -0,0 +1,235 @@
...
+			strings.ReplaceAll(devInfo.Model, " ", "-"),
+			endpointRefAddr)
+
+		discovered = sdkModel.DiscoveredDevice{
+			Name:        deviceName,
+			Protocols:   dev.Protocols,
+			Description: fmt.Sprintf("%s %s Camera", devInfo.Manufacturer, devInfo.Model),
+			Labels:      []string{"auto-discovery", devInfo.Manufacturer, devInfo.Model},
+		}
+		d.lc.Debugf("Discovered camera from the address '%s'", xaddr)
```
lenny-goodell: Include name?

internal/driver/onvifdiscovery.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	return discovered, nil
+}
+
+// mapProbeResults converts a slice of discovered onvif.Device into the generic
+// netscan.ProbeResult.
+func mapProbeResults(host, port string, devices []onvif.Device) (res []netscan.ProbeResult) {
+	for _, dev := range devices {
```
lenny-goodell:
```suggestion
	for _, device := range devices {
```

internal/driver/onvifdiscovery.go
```diff
@@ -0,0 +1,235 @@
...
+			if !errors.Is(err, os.ErrDeadlineExceeded) {
+				params.Logger.Debugf("Unexpected error occurred while reading ws-discovery responses: %s", err.Error())
+			}
+			break
+		}
+		responses = append(responses, string(buf[0:n]))
+	}
+
+	if len(responses) == 0 {
+		params.Logger.Tracef("%s: No Response", addr)
```
lenny-goodell, line 191:
```suggestion
		params.Logger.Debugf("%s: No Response", addr)
```

internal/driver/onvifdiscovery.go
```diff
@@ -0,0 +1,235 @@
...
+//			d.lc.Warnf("Registered device %s is missing required %s protocol information: EndpointRefAddress.",
+//				dev.Name, OnvifProtocol)
+//			continue
+//		}
+//
+//		deviceMap[endpointRef] = dev
+//	}
+//
+//	return deviceMap
+//}
```
lenny-goodell: Remove commented out code

pkg/netscan/discover.go
```diff
@@ -0,0 +1,271 @@
...
+	NetworkUDP = "udp"
+	NetworkTCP = "tcp"
+)
+
+// AutoDiscover probes all addresses in the configured network to attempt to discover any possible
+// devices for a specific protocol
+func AutoDiscover(ctx context.Context, proto ProtocolSpecificDiscovery, params Params) []DiscoveredDevice {
+	params.Logger.Debugf("AutoDiscover called with the following parameters: %+v", params)
+	if len(params.Subnets) == 0 {
+		params.Logger.Warnf("Discover was called, but no subnet information has been configured!")
```
lenny-goodell:
```suggestion
		params.Logger.Warn("Discover was called, but no subnet information has been configured!")
```

pkg/netscan/discover.go
```diff
@@ -0,0 +1,271 @@
...
+
+		ipnets = append(ipnets, ipnet)
+		// compute the estimate total amount of network probes we are going to make
+		// this is an estimate because it may be lower due to skipped addresses (existing devices)
+		sz, _ := ipnet.Mask.Size()
+		estimatedProbes += int(computeNetSz(sz))
+	}
+
+	if estimatedProbes == 0 {
+		params.Logger.Warnf("No valid CIDRs provided, unable to scan for devices.")
```
lenny-goodell:
```suggestion
		params.Logger.Warn("No valid CIDRs provided, unable to scan for devices.")
```

pkg/netscan/discover.go
```diff
@@ -0,0 +1,271 @@
...
+// process any in-flight results.
+func processResultChannel(resultCh chan []ProbeResult, proto ProtocolSpecificDiscovery, params Params) []DiscoveredDevice {
+	devices := make([]DiscoveredDevice, 0)
+	for probeResults := range resultCh {
+		if len(probeResults) == 0 {
+			continue
+		}
+
+		for _, probeResult := range probeResults {
+			dev, err := proto.ConvertProbeResult(probeResult, params)
```
lenny-goodell:
```suggestion
			device, err := proto.ConvertProbeResult(probeResult, params)
```

pkg/netscan/discover.go
```diff
@@ -0,0 +1,271 @@
...
+
+func handleConnectionInternal(host string, port string, conn net.Conn, params workerParams) {
+	// on udp, the dial is always successful, so don't print
+	if !strings.HasPrefix(params.NetworkProtocol, NetworkUDP) {
+		params.Logger.Debugf("Connection dialed %s://%s:%s", params.NetworkProtocol, host, port)
+	}
+
+	results, err := params.proto.OnConnectionDialed(host, port, conn, params.Params)
+	if err != nil {
+		params.Logger.Debugf(err.Error())
```
lenny-goodell: Comment why Debug and Not Error or Warn log message

```suggestion
		params.Logger.Debug(err.Error())
```

pkg/netscan/discover.go
```diff
@@ -0,0 +1,271 @@
...
+func ipWorker(params workerParams) {
+	ip := net.IP([]byte{0, 0, 0, 0})
+
+	for {
+		select {
+		case <-params.ctx.Done():
+			// stop working if we have been cancelled
+			return
+
+		case a, ok := <-params.ipCh:
```
lenny-goodell: Use better name for `a`

lenny-goodell (changes requested):
Completed reviewing all files

cloudxxx8:
> > Btw, the reason I have organized the code as such (adding a sub go-module for netscan vs just a go package) is that the code is based on the LLRP discovery code I wrote, and the intention would be for both of them to share it (whether that means it gets its own repo, or what). I
> 
> @ajcasagrande , @cloudxxx8 , @weichou1229 , should this `netscan` code be integrated in to the Device SDK, rather than embedded in this repo?
> 
> It will need to be in this repo until we release this service for Kamakura. After that it should be integrated in the Device SDK.

@iain-anderson for awareness
if this logic can be reused by multiple device services, I believe we can move it to the SDK in the next release.

cmd/res/configuration.toml
ajcasagrande: Agreed, and we will likely migrate it all in another sprint/story.

internal/driver/driver.go
ajcasagrande: I agree, `device` is a better name, but it does conflict with imported package name `"github.com/IOTechSystems/onvif/device"`, which is why I changed it.

Screenshot from 2022-05-27 10-29-28

internal/driver/onvifdiscovery.go
ajcasagrande: The reason I chose trace instead of debug is that when using UDP there is no handshake so no response is returned for every single IP address that does not respond, meaning you can have up to 254 logs in a standard `/24` subnet.

The more interesting debug logs look like so: `msg="read udp 192.168.0.190:41133->192.168.0.1:3702: recvfrom: connection refused"` meaning the host is online but refusing the connection.

internal/driver/onvifdiscovery.go
ajcasagrande: only commented out to keep the linter happy. the code is in use in the rediscovery PR. Now that it is being used there I can remove it here and not lose it.

pkg/netscan/discover.go
ajcasagrande: That function doesn't exist on the interface. But if I switch to EdgeX logger I will remove the `f`.

pkg/netscan/discover.go
ajcasagrande: That function doesn't exist on the interface. But if I switch to EdgeX logger I will remove the `f`.

doc/auto-discovery-docker.md
ajcasagrande: Not sure what you are referring to here?

internal/driver/service_test.go
ajcasagrande: This is legacy code from the LLRP repo that never got migrated to use mockery. I don't think we are actually using this mock yet, so maybe I will remove it for now.

internal/driver/service.go
ajcasagrande: https://github.com/edgexfoundry/device-sdk-go/issues/1168

pkg/netscan/types.go
ajcasagrande: The original intent was to create a platform agnostic solution. Currently the netscan package has no EdgeX references and can be used as a standalone go module. The code can be "EdgeX-ified" now, especially if the intent is to make it native sdk code.

internal/driver/onvifdiscovery.go
ajcasagrande: This is kind of a hack that allows you to pre-fill the credentials per camera by using the endpoint ref. The upcoming story I am working on would replace this, so maybe I should just get rid of it for now? Or should I keep and just comment?

pkg/netscan/discover.go
ajcasagrande: Will comment about why debug.

doc/auto-discovery-docker.md
ajcasagrande: A lot of this was changed before that PR happened. I think any comments against this section can be handled as part of #69

doc/auto-discovery-docker.md
ajcasagrande: Same comment, will be addressed as part of the fix for #69

internal/driver/driver.go
ajcasagrande: I can change them all back to `device`, but maybe we should consider using an import alias for that package?

ajcasagrande:
> > @ajcasagrande , @cloudxxx8 , @weichou1229 , should this `netscan` code be integrated in to the Device SDK, rather than embedded in this repo?
> > It will need to be in this repo until we release this service for Kamakura. After that it should be integrated in the Device SDK.
> 
> @iain-anderson for awareness if this logic can be reused by multiple device services, I believe we can move it to the SDK in the next release.

I agree that it makes sense that the netscan module can be "EdgeX-ified" and moved into the Device SDK itself.

internal/driver/service_test.go
ajcasagrande: removed

Dockerfile
```diff
@@ -28,7 +28,7 @@ RUN apk add --no-cache ${ALPINE_PKG_BASE} ${ALPINE_PKG_EXTRA}
 
 WORKDIR /device-onvif-camera
 
-COPY go.mod vendor* ./
+COPY go.mod vendor* pkg/netscan/go.mod ./
```
weichou1229: Where is the file `netscan/go.mod`?

internal/driver/onvifdiscovery.go
```diff
@@ -0,0 +1,204 @@
...
+// executeRawProbe essentially performs a UDP unicast ws-discovery probe by sending the
+// probe message directly over the connection and listening for any responses. Those
+// responses are then converted into a slice of onvif.Device.
+func executeRawProbe(conn net.Conn, params netscan.Params) ([]onvif.Device, error) {
+	probeSOAP := wsdiscovery.BuildProbeMessage(uuid.Must(uuid.NewV4()).String(), nil, nil,
+		map[string]string{"dn": "http://www.onvif.org/ver10/network/wsdl"})
+
+	addr := conn.RemoteAddr().String()
+	if err := conn.SetDeadline(time.Now().Add(params.Timeout)); err != nil {
+		return nil, errors.Wrapf(err, "%s: failed to set read/write deadline", addr)
```
weichou1229: Can we use EdgeX's errors package instead of pkg/errors?

Dockerfile
ajcasagrande: has been removed now. will revert this. good catch.

internal/driver/onvifdiscovery.go
ajcasagrande: yeah, let me convert this.

doc/auto-discovery-docker.md
lenny-goodell: `Export the Security Bootstrapper`

internal/driver/driver.go
lenny-goodell: Yes, prefer an alias for the package in these cases.

internal/driver/service.go
lenny-goodell: Thanks!

internal/driver/onvifdiscovery.go
lenny-goodell: Either works for me.

internal/driver/onvifdiscovery.go
lenny-goodell: ok, makes sense

lenny-goodell:
> this `netscan` code be integrated in to the Device SDK, rather than embedded in this repo?

@ajcasagrande , please submit issue on device-sdk-go to migrate your `netscan` code into the SDK. THX!

doc/auto-discovery-docker.md
ajcasagrande: In that case I do not know. Is this something that can be addressed as part of #69 ?

lenny-goodell: I think this was for running hybrid mode??  Compose builder will generate the compose file so this isn't needed, so yea it can be part of #69. I'll add it to the description.

lenny-goodell (dismissed):
LGTM

internal/driver/driver.go
```diff
@@ -303,18 +315,29 @@ func (d *Driver) createOnvifClient(deviceName string) error {
...
+		Username: secretData[secret.UsernameKey],
+		Password: secretData[secret.PasswordKey],
+	}, nil
+}
 
+// getCredentials will repeatedly try and get the credentials located at secretPath from
+// secret provider every CredentialsRetryTime seconds for a maximum of CredentialsRetryWait seconds.
+// Note that this function will block until either the credentials are found, or CredentialsRetryWait
+// seconds have elapsed.
+func (d *Driver) getCredentials(secretPath string) (credentials config.Credentials, err errors.EdgeX) {
```
lenny-goodell, line 335: @cloudxxx8,  We tend to be doing this is all our device services. I am thinking this retry needs to be built into the SecretProvider. Something like GetSecretWithRetry(...). I will create issues.

lenny-goodell (approved):
LGTM

cloudxxx8 (commented):
@ajcasagrande this pr add a new import, but it is not listed in the PR description
github.com/gofrs/uuid v3.2.0+incompatible

this should not be a necessary import.  please help remove it.

ajcasagrande:
@cloudxxx8 yes, I can create a PR to remove.

PR #87, opened 2022-06-21
Custom metadata

ChristianDarr-personal:
## PR Checklist

- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

- Device service only has ONVIF Protocols as a way of storing extra metadata.

## What is the new behavior?
- A new CustomMetadata resource has been added to get and set new protocol property map Custom Metadata

- [ ] Yes
- [x] No

## New Imports
No

## Specific Instructions
The device profile must be updated to include the resource.

## Other information

### DIFF
--- a/cmd/res/profiles/camera.yaml
+++ b/cmd/res/profiles/camera.yaml
@@ -720,3 +720,14 @@ deviceCommands:
     resourceOperations:
       - { deviceResource: "Profiles" }
       - { deviceResource: "StreamUri" }
+
+ - name: "CustomMetadata"
+    isHidden: false
+    description: "Update or query customized metadata configurations"
+    attributes:
+      service: "EdgeX"
+      setFunction:
+      getFunction:
+    properties:
+      valueType: "Object"
+      readWrite: "RW"
--- a/cmd/res/profiles/camera.yaml
+++ b/cmd/res/profiles/camera.yaml
@@ -162,6 +162,16 @@ deviceResources:
     properties:
       valueType: "Object"
       readWrite: "W"
+  - name: "CustomMetadata"
+    isHidden: false
+    description: "Return the options for the supported rules that specify an Option attribute."
+    attributes:
+      service: "Device"
+      getFunction: "GetCustomMetadata"
+      setFunction: "SetCustomMetadata"
+    properties:
+      valueType: "Object"
+      readWrite: "RW"
 
   # Video Streaming
   - name: "Profiles"
@@ -720,14 +730,3 @@ deviceCommands:
     resourceOperations:
       - { deviceResource: "Profiles" }
       - { deviceResource: "StreamUri" }
-
- - name: "CustomMetadata"
-    isHidden: false
-    description: "Update or query customized metadata configurations"
-    attributes:
-      service: "EdgeX"
-      setFunction:
-      getFunction:
-    properties:
-      valueType: "Object"
-      readWrite: "RW"
--- a/cmd/res/profiles/camera.yaml
+++ b/cmd/res/profiles/camera.yaml
@@ -172,6 +172,15 @@ deviceResources:
     properties:
       valueType: "Object"
       readWrite: "RW"
+  - name: "DeleteCustomMetadata"
+    isHidden: false
+    description: "Delete custom metadata for the camera"
+    attributes:
+      service: "EdgeX"
+      setFunction: "DeleteCustomMetadata"
+    properties:
+      valueType: "Object"
+      readWrite: "W"
 
   # Video Streaming
   - name: "Profiles"
--- a/cmd/res/profiles/camera.yaml
+++ b/cmd/res/profiles/camera.yaml
@@ -164,7 +164,7 @@ deviceResources:
       readWrite: "W"
   - name: "CustomMetadata"
     isHidden: false
-    description: "Return the options for the supported rules that specify an Option attribute."
+    description: "Get and set custom added information about the cameras"
     attributes:
       service: "Device"
       custom: "Custom"
--- a/cmd/res/profiles/camera.yaml
+++ b/cmd/res/profiles/camera.yaml
@@ -167,6 +167,7 @@ deviceResources:
     description: "Return the options for the supported rules that specify an Option attribute."
     attributes:
       service: "Device"
+      custom: "Custom"
       getFunction: "GetCustomMetadata"
       setFunction: "SetCustomMetadata"
     properties:
--- a/cmd/res/profiles/camera.yaml
+++ b/cmd/res/profiles/camera.yaml
@@ -164,10 +164,9 @@ deviceResources:
       readWrite: "W"
   - name: "CustomMetadata"
     isHidden: false
-    description: "Get and set custom added information about the cameras"
+    description: "Get and set custom metadata for the camera"
     attributes:
-      service: "Device"
-      custom: "Custom"
+      service: "EdgeX"
       getFunction: "GetCustomMetadata"
       setFunction: "SetCustomMetadata"
     properties:
--- a/doc/custom-metadata-feature.md
+++ b/doc/custom-metadata-feature.md
@@ -62,7 +62,7 @@ Use the CustomMetadata resource to get and display the fields of *CustomMetadata
 1. Use this command to return all of the data in the CustomMetadata field.
 
 ```shell
-curl http://localhost://59882/api/v2/device/name/<device name>/CustomMetadata | json_pp
+curl http://localhost:59882/api/v2/device/name/<device name>/CustomMetadata | json_pp
 ```
 2. The repsonse from the curl command.
 ```shell
--- a/doc/custom-metadata-feature.md
+++ b/doc/custom-metadata-feature.md
@@ -1,0 +1,31 @@
+# Custom Metadata
+
+We have added a custom device resource for setting and querying custom metadata for each camera added to the service.
+
+## Usage
+
+- The *CustomMetadata* map is an element in the *ProtocolProperties* device field. It is initialized to be empty, so the user can add their desired fields.
+
+### Set Custom Metadata
+
+Use the CustomMetadata resource to set the fields of *CustomMetadata*. Choose the key/value pairs to represent your custom fields.
+
+```shell
+curl -X POST http://localhost:59882/api/v2/device/name/<device name>/CustomMetadata /
+-H 'Content-Type: application/json'/
+-d '{
+        "CustomMetadata: {
+            "CommonName":"Front Door Camera",
+            "Location":"Front Door"
+        }
+    }
+```
+>Example of SetCustomMetadata command.
+
+### Get Custom Metadata
+
+Use the CustomMetadata resource to get and display the fields of *CustomMetadata*.
+
+```shell
+curl http://localhost://59882/api/v2/device/name/<device name>/CustomMetadata
+```
--- a/doc/custom-metadata-feature.md
+++ b/doc/custom-metadata-feature.md
@@ -10,22 +10,159 @@ We have added a custom device resource for setting and querying custom metadata 
 
 Use the CustomMetadata resource to set the fields of *CustomMetadata*. Choose the key/value pairs to represent your custom fields.
 
+1. Use this command to put the data in the CustomMetadata field.
 ```shell
-curl -X POST http://localhost:59882/api/v2/device/name/<device name>/CustomMetadata /
--H 'Content-Type: application/json'/
--d '{
-        "CustomMetadata: {
+curl --request PUT 'http://0.0.0.0:59882/api/v2/device/name/<device name>/CustomMetadata' \
+    --header 'Content-Type: application/json' \
+    --data-raw '{
+        "CustomMetadata": {
             "CommonName":"Front Door Camera",
-            "Location":"Front Door"
+            "Location":"Front Door",
+            "Color":"Black and white",
+            "Condition": "Good working condition"
         }
-    }
+    }' | json_pp
 ```
->Example of SetCustomMetadata command.
+2. The response from the curl command.
+```
+{
+    "apiVersion": "v2",
+    "statusCode": 200
+}
+```
+
 
 ### Get Custom Metadata
 
 Use the CustomMetadata resource to get and display the fields of *CustomMetadata*.
 
+1. Use this command to return all of the data in the CustomMetadata field.
+
 ```shell
-curl http://localhost://59882/api/v2/device/name/<device name>/CustomMetadata
+curl http://localhost://59882/api/v2/device/name/<device name>/CustomMetadata | json_pp
+```
+2. The repsonse from the curl command.
+```shell
+{
+   "apiVersion" : "v2",
+   "event" : {
+      "apiVersion" : "v2",
+      "deviceName" : "3fa1fe68-b915-4053-a3e1-cc32e5000688",
+      "id" : "ba3987f9-b45b-480a-b582-f5501d673c4d",
+      "origin" : 1655409814077374935,
+      "profileName" : "onvif-camera",
+      "readings" : [
+         {
+            "deviceName" : "3fa1fe68-b915-4053-a3e1-cc32e5000688",
+            "id" : "cf96e5c0-bde1-4c0b-9fa4-8f765c8be456",
+            "objectValue" : {
+               "Color" : "Black and white",
+               "CommonName" : "Front Door Camera",
+               "Condition" : "Good working condition",
+               "Location" : "Front Door"
+            },
+            "origin" : 1655409814077374935,
+            "profileName" : "onvif-camera",
+            "resourceName" : "CustomMetadata",
+            "value" : "",
+            "valueType" : "Object"
+         }
+      ],
+      "sourceName" : "CustomMetadata"
+   },
+   "statusCode" : 200
+}
+```
+
+
+
+### Get Specific Custom Metadata
+
+Pass the *CustomMetadata* resource a query to get specific field(s) in CustomMetadata. The query must be a base64 encoded json object with an array of fields you want to access.
+
+1. Json object holding an array of fields you want to query.
+```json
+'{
+    "CustomMetadata": 
+        [
+            "CommonName",
+            "Location"
+        ]
+}'
+```
+
+2. Use this command to convert the json object to base64.
+```shell
+echo '{
+    "CustomMetadata": 
+        [
+            "CommonName",
+            "Location"
+        ]
+}' | base64
+```
+
+3. The response converted to base64.
+```shell
+ewogICAgIkN1c3RvbU1ldGFkYXRhIjogCiAgICAgICAgWwogICAgICAgICAgICAiQ29tbW9uTmFtZSIsCiAgICAgICAgICAgICJMb2NhdGlvbiIKICAgICAgICBdCn0K
+```
+
+4. Use this command to query the fields you provided in the json object.
+```shell
+curl http://localhost:59882/api/v2/device/name/3fa1fe68-b915-4053-a3e1-cc32e5000688/CustomMetadata?jsonObject=ewogICAgIkN1c3RvbU1ldGFkYXRhIjogCiAgICAgICAgWwogICAgICAgICAgICAiQ29tbW9uTmFtZSIsCiAgICAgICAgICAgICJMb2NhdGlvbiIKICAgICAgICBdCn0K | json_pp
+
+```
+
+5. Curl response. 
+```shell
+{
+   "apiVersion" : "v2",
+   "event" : {
+      "apiVersion" : "v2",
+      "deviceName" : "3fa1fe68-b915-4053-a3e1-cc32e5000688",
+      "id" : "24c3eb0a-48b1-4afe-b874-965aeb2e42a2",
+      "origin" : 1655410556448058195,
+      "profileName" : "onvif-camera",
+      "readings" : [
+         {
+            "deviceName" : "3fa1fe68-b915-4053-a3e1-cc32e5000688",
+            "id" : "d0c26303-20b5-4ccd-9e63-fb02b87b8ebc",
+            "objectValue" : {
+               "CommonName" : "Front Door Camera",
+               "Location" : "Front Door"
+            },
+            "origin" : 1655410556448058195,
+            "profileName" : "onvif-camera",
+            "resourceName" : "CustomMetadata",
+            "value" : "",
+            "valueType" : "Object"
+         }
+      ],
+      "sourceName" : "CustomMetadata"
+   },
+   "statusCode" : 200
+}
+```
+
+### Additional Usage
+
+Setting a field to be empty will delete that field.
+
+1. Use this command to delete fields.
+```shell
+curl --request PUT 'http://0.0.0.0:59882/api/v2/device/name/<device name>/CustomMetadata' \
+    --header 'Content-Type: application/json' \
+    --data-raw '{
+        "CustomMetadata": {
+            "Color":"",
+            "Condition": ""
+        }
+    }' | json_pp
+```
+2. The response from the curl command.
+```
+{
+    "apiVersion": "v2",
+    "statusCode": 200
+}
 ```
--- a/doc/custom-metadata-feature.md
+++ b/doc/custom-metadata-feature.md
@@ -1,10 +1,10 @@
 # Custom Metadata
 
-We have added a custom device resource for setting and querying custom metadata for each camera added to the service.
+Custom metadata can be applied and retreived for each camera added to the service.
 
 ## Usage
 
-- The *CustomMetadata* map is an element in the *ProtocolProperties* device field. It is initialized to be empty, so the user can add their desired fields.
+- The `CustomMetadata` map is an element in the `ProtocolProperties` device field. It is initialized to be empty, so the user can add their desired fields.
 
 ### Preset Custom Metadata
 
@@ -31,7 +31,7 @@ Description = "onvif conformant camera"
 
 ### Set Custom Metadata
 
-Use the CustomMetadata resource to set the fields of *CustomMetadata*. Choose the key/value pairs to represent your custom fields.
+Use the CustomMetadata resource to set the fields of `CustomMetadata`. Choose the key/value pairs to represent your custom fields.
 
 1. Use this command to put the data in the CustomMetadata field.
 ```shell
@@ -57,7 +57,7 @@ curl --request PUT 'http://0.0.0.0:59882/api/v2/device/name/<device name>/Custom
 
 ### Get Custom Metadata
 
-Use the CustomMetadata resource to get and display the fields of *CustomMetadata*.
+Use the CustomMetadata resource to get and display the fields of `CustomMetadata`.
 
 1. Use this command to return all of the data in the CustomMetadata field.
 
@@ -101,7 +101,7 @@ curl http://localhost:59882/api/v2/device/name/<device name>/CustomMetadata | js
 
 ### Get Specific Custom Metadata
 
-Pass the *CustomMetadata* resource a query to get specific field(s) in CustomMetadata. The query must be a base64 encoded json object with an array of fields you want to access.
+Pass the `CustomMetadata` resource a query to get specific field(s) in CustomMetadata. The query must be a base64 encoded json object with an array of fields you want to access.
 
 1. Json object holding an array of fields you want to query.
 ```json
--- a/doc/custom-metadata-feature.md
+++ b/doc/custom-metadata-feature.md
@@ -169,7 +169,7 @@ curl http://localhost:59882/api/v2/device/name/3fa1fe68-b915-4053-a3e1-cc32e5000
 
 ### Additional Usage
 
-Setting a field to be empty will delete that field.
+Setting a field to "deletes" will delete that field.
 
 1. Use this command to delete fields.
 ```shell
@@ -177,8 +177,8 @@ curl --request PUT 'http://0.0.0.0:59882/api/v2/device/name/<device name>/Custom
     --header 'Content-Type: application/json' \
     --data-raw '{
         "CustomMetadata": {
-            "Color":"",
-            "Condition": ""
+            "Color":"delete",
+            "Condition": "delete"
         }
     }' | json_pp
 ```
--- a/doc/custom-metadata-feature.md
+++ b/doc/custom-metadata-feature.md
@@ -53,6 +53,7 @@ curl --request PUT 'http://0.0.0.0:59882/api/v2/device/name/<device name>/Custom
     "statusCode": 200
 }
 ```
+>Note: ensure all data is properly formatted json, and that all special characters are escaped if necessary
 
 
 ### Get Custom Metadata
--- a/doc/custom-metadata-feature.md
+++ b/doc/custom-metadata-feature.md
@@ -6,6 +6,29 @@ We have added a custom device resource for setting and querying custom metadata 
 
 - The *CustomMetadata* map is an element in the *ProtocolProperties* device field. It is initialized to be empty, so the user can add their desired fields.
 
+### Preset Custom Metadata
+
+If you add pre-defined devices, set up the CustomMetadata object as shown in the camera.toml.example file.
+
+```toml
+# Pre-defined Devices
+[[DeviceList]]
+Name = "Camera001"
+ProfileName = "onvif-camera"
+Description = "onvif conformant camera"
+  [DeviceList.Protocols]
+    [DeviceList.Protocols.Onvif]
+    Address = "192.168.12.123"
+    Port = "80"
+    # Assign AuthMode to "usernametoken" | "digest" | "both" | "none"
+    AuthMode = "digest"
+    SecretPath = "credentials001"
+    [DeviceList.Protocols.CustomMetadata]
+    CommonName = "Door camera"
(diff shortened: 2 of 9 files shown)


weichou1229:
@ChristianDarr-personal I thought the `CustomMetadata` is not appropriate for the protocol properties because the protocol properties are usually used to describe the connection info for the specified device. 

@ajcasagrande And it worth raising a requirement on the core working group to add a new field like `Tags map[string]interface{}`?

lenny-goodell:
@weichou1229 , I think that may be something to look at post release. Since we are targeting `Kamakura` for the first release, we are limited to the protocols section.

cmd/res/profiles/camera.yaml
```diff
@@ -162,6 +162,17 @@ deviceResources:
     properties:
       valueType: "Object"
       readWrite: "W"
+  - name: "CustomMetadata"
+    isHidden: false
+    description: "Return the options for the supported rules that specify an Option attribute."
```
ajcasagrande: description makes no sense

doc/custom-metadata-feature.md
```diff
@@ -0,0 +1,191 @@
+# Custom Metadata
+
+We have added a custom device resource for setting and querying custom metadata for each camera added to the service.
```
ajcasagrande: instead of reading like this, it should sound more like: "Custom metadata can be applied and retrieved for each camera registered in the system"

doc/custom-metadata-feature.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Usage
+
+- The *CustomMetadata* map is an element in the *ProtocolProperties* device field. It is initialized to be empty, so the user can add their desired fields.
```
ajcasagrande:
```suggestion
- The `CustomMetadata` map is an element in the `ProtocolProperties` device field. It is initialized to be empty, so the user can add their desired fields.
```

doc/custom-metadata-feature.md
```diff
@@ -0,0 +1,191 @@
...
+    SecretPath = "credentials001"
+    [DeviceList.Protocols.CustomMetadata]
+    CommonName = "Door camera"
+    Location = "Front door"
+```
+
+
+### Set Custom Metadata
+
+Use the CustomMetadata resource to set the fields of *CustomMetadata*. Choose the key/value pairs to represent your custom fields.
```
ajcasagrande: Like above, instead of using `*CustomMetadata*`, you should use backticks ``` `CustomMetadata` ```

doc/custom-metadata-feature.md
```diff
@@ -0,0 +1,191 @@
...
+```shell
+curl --request PUT 'http://0.0.0.0:59882/api/v2/device/name/<device name>/CustomMetadata' \
+    --header 'Content-Type: application/json' \
+    --data-raw '{
+        "CustomMetadata": {
+            "Color":"delete",
+            "Condition": "delete"
+        }
+    }' | json_pp
+```
```
ajcasagrande: I think we should find a better way to handle this. Each of them have some drawbacks or hackiness to them, but here are some alternatives:
1. Having a write-only `DeleteCustomMetadata` resource that takes the fields you want to delete as the json body
     1. Problem with this is that this `DeleteCustomMetadata` "resource" isnt really a "resource " so to speak.

2. Modifying the input object to have a special `DeleteMetadata` field (or other name) that you pass in a list of fields to delete.
     1. Depending on how it is possible may have to use csv string instead of array

3.  Keep it the same, but make the `delete` keyword less prone to collisions? `_DELETE_`? or something like that?

@trleasher-intel @lenny-intel thoughts?

internal/driver/custommetadata.go
```diff
@@ -0,0 +1,84 @@
...
+// setCustomMetadata will return a map containing the fields provided in the call to the function
+func (onvifClient *OnvifClient) setCustomMetadata(device contract.Device, data []byte) (contract.Device, errors.EdgeX) {
+	var dataObj contract.ProtocolProperties
+
+	err := json.Unmarshal(data, &dataObj)
+	if err != nil {
+		return device, errors.NewCommonEdgeX(errors.KindServerError, "failed to unmarshal the json request body", err)
+	}
+	if len(dataObj) == 0 {
+		return device, errors.NewCommonEdgeX(errors.KindServerError, "no data in PUT command", err)
```
ajcasagrande: this should probably be an errors.KindXXX that more relates to bad user input.

internal/driver/custommetadata.go
```diff
@@ -0,0 +1,84 @@
...
+		trimValue := strings.TrimSpace(value)
+		trimKey := strings.TrimSpace(key)
+		if trimValue == "delete" || trimValue == "Delete" {
+			delete(device.Protocols[CustomMetadata], trimKey)
+			continue
+		}
+		if len(trimKey) == 0 {
+			continue
+		}
+		device.Protocols[CustomMetadata][trimKey] = trimValue // create or update a field in CustomMetadata
```
ajcasagrande: should we check if `CustomMetadata` exists inside `device.Protocols` first?

internal/driver/custommetadata.go
```diff
@@ -0,0 +1,84 @@
...
+	err := json.Unmarshal(data, &dataObj)
+	if err != nil {
+		return device, errors.NewCommonEdgeX(errors.KindServerError, "failed to unmarshal the json request body", err)
+	}
+	if len(dataObj) == 0 {
+		return device, errors.NewCommonEdgeX(errors.KindServerError, "no data in PUT command", err)
+	}
+	for key, value := range dataObj {
+		trimValue := strings.TrimSpace(value)
+		trimKey := strings.TrimSpace(key)
```
ajcasagrande: if never using the non-trimmed versions, just overwrite them with the same name
```suggestion
		value = strings.TrimSpace(value)
		key = strings.TrimSpace(key)
```

internal/driver/custommetadata.go
```diff
@@ -0,0 +1,84 @@
...
+		if trimValue == "delete" || trimValue == "Delete" {
+			delete(device.Protocols[CustomMetadata], trimKey)
+			continue
+		}
+		if len(trimKey) == 0 {
+			continue
+		}
```
ajcasagrande: this should probably go before the delete check

internal/driver/custommetadata.go
```diff
@@ -0,0 +1,84 @@
...
+		device.Protocols[CustomMetadata][trimKey] = trimValue // create or update a field in CustomMetadata
+	}
+
+	return device, nil
```
ajcasagrande, line 49: this code may suffer from race conditions on the device object, but that may be a limitation of the sdk and need to be enhanced there or we refactor this whole service to add locking on the devices.

internal/driver/custommetadata.go
```diff
@@ -0,0 +1,84 @@
...
+		metadataObj = device.Protocols[CustomMetadata]
+	} else { // if a list of fields is provided, return those specific fields
+		metadataObj, err = onvifClient.getSpecificCustomMetadata(device, data)
+		if err != nil {
+			return nil, errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf("failed to get specific metadata for device %s", onvifClient.DeviceName), err)
+		}
+	}
+
+	return metadataObj, nil
+}
```
ajcasagrande: I think this can be simplified

```suggestion
func (onvifClient *OnvifClient) getCustomMetadata(device contract.Device, data []byte) (contract.ProtocolProperties, errors.EdgeX) {
	if len(data) == 0 { // if no list is provided, return all
		return device.Protocols[CustomMetadata], nil
	}

	// if a list of fields is provided, return those specific fields
	metadataObj, err := onvifClient.getSpecificCustomMetadata(device, data)
	if err != nil {
		return nil, errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf("failed to get specific metadata for device %s", onvifClient.DeviceName), err)
	}
	return metadataObj, nil
}
```

internal/driver/custommetadata.go
```diff
@@ -0,0 +1,84 @@
...
+func (onvifClient *OnvifClient) getSpecificCustomMetadata(device contract.Device, data []byte) (obj contract.ProtocolProperties, error errors.EdgeX) {
+	dataArray := make(map[string][]string)
+	dataMap := make(map[string]string)
+
+	err := json.Unmarshal(data, &dataArray)
+	if err != nil {
+		return nil, errors.NewCommonEdgeX(errors.KindServerError, "failed to unmarshal the json request body", err)
+	}
+	if len(dataArray[CustomMetadata]) == 0 {
+		return nil, errors.NewCommonEdgeX(errors.KindServerError, "no data in query body", err)
```
ajcasagrande: user input error

internal/driver/custommetadata.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	for _, key := range dataArray[CustomMetadata] {
+		value := device.Protocols[CustomMetadata][key]
+		if value == "" {
```
ajcasagrande: Should this be changed to `value, found := ...; if !found`? Because in theory a key can exist with an empty string value

internal/driver/custommetadata.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+	}
+
+	return metadataObj, nil
+}
+
+// getSpecificCustomMetadata will return a map of the key/value pairs corresponding to the array of keys provided in the resource call
+func (onvifClient *OnvifClient) getSpecificCustomMetadata(device contract.Device, data []byte) (obj contract.ProtocolProperties, error errors.EdgeX) {
+	dataArray := make(map[string][]string)
+	dataMap := make(map[string]string)
```
ajcasagrande: i would rename this to something like `response`

internal/driver/custommetadata.go
```diff
@@ -0,0 +1,84 @@
...
+			return nil, errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf("failed to get specific metadata for device %s", onvifClient.DeviceName), err)
+		}
+	}
+
+	return metadataObj, nil
+}
+
+// getSpecificCustomMetadata will return a map of the key/value pairs corresponding to the array of keys provided in the resource call
+func (onvifClient *OnvifClient) getSpecificCustomMetadata(device contract.Device, data []byte) (obj contract.ProtocolProperties, error errors.EdgeX) {
+	dataArray := make(map[string][]string)
```
ajcasagrande: i would rename this to something like `input`, `jsonInput`, `request`, etc.

internal/driver/custommetadata_test.go
```diff
@@ -0,0 +1,277 @@
...
+		errorExpected bool
+	}{
+		{"happy path without data - getAll", myDevice, "", myDevice.Protocols[CustomMetadata], false},
+		{"happy path with data for single field", myDevice, singleDataTest.data, singleDataTest.expectedResponse, false},
+		{"happy path with data for multiple field", myDevice, happyMultipleDataTest.data, happyMultipleDataTest.expectedResponse, false},
+		{"happy path with data for non-existent field", myDevice, noFieldDataTest.data, noFieldDataTest.expectedResponse, false},
+		{"happy path with data for multiple non-existent fields", myDevice, noFieldsDataTest.data, noFieldsDataTest.expectedResponse, false},
+		{"with empty data", myDevice, emptyData.data, emptyData.expectedResponse, true},
+		{"badJson", myDevice, "bogus", myDevice.Protocols[CustomMetadata], true},
+	}
```
ajcasagrande: For all of these test cases, you should refactor them such that there is only one test struct creation and the individual test cases are not created above using variables, but instead as inlined values as part of the tests. 

So your `tests := []struct {` and your `type testCase struct` structs should be merged into 1 struct, with all the tests being defined inline in one place.

internal/driver/custommetadata_test.go
```diff
@@ -0,0 +1,277 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
ajcasagrande: I usually call this `test`, and also you should get in the habit now of defining a temp variable override with the same name (or different name) in your table tests to avoid issues when called in parallel:

```
for _, test := range tests {
		test := test   <- this is save the state of the `test` var when running in parallel
		t.Run(test.name, func(t *testing.T) {
```

ajcasagrande (changes requested):
Have 1 or 2 files left to check, submitting feedback so far...

cmd/res/profiles/camera.yaml
```diff
@@ -164,7 +164,7 @@ deviceResources:
       readWrite: "W"
   - name: "CustomMetadata"
     isHidden: false
-    description: "Return the options for the supported rules that specify an Option attribute."
+    description: "Get and set custom added information about the cameras"
```
ajcasagrande:
```suggestion
    description: "Get and set custom metadata about the camera"
```

internal/driver/custommetadata.go
```diff
@@ -24,19 +24,24 @@ func (onvifClient *OnvifClient) setCustomMetadata(device contract.Device, data [
 		return device, errors.NewCommonEdgeX(errors.KindServerError, "failed to unmarshal the json request body", err)
 	}
 	if len(dataObj) == 0 {
-		return device, errors.NewCommonEdgeX(errors.KindServerError, "no data in PUT command", err)
+		return device, errors.NewCommonEdgeX(errors.KindIOError, "no data in PUT command", err)
```
ajcasagrande: if this was http, I would expect a 400 bad request, so according to edgex, that would map to `errors.KindContractInvalid`

```go
	case http.StatusBadRequest:
		return KindContractInvalid
```

internal/driver/custommetadata.go
```diff
@@ -24,19 +24,24 @@ func (onvifClient *OnvifClient) setCustomMetadata(device contract.Device, data [
...
-		if len(trimKey) == 0 {
+		if value == "delete" || value == "Delete" {
+			delete(device.Protocols[CustomMetadata], key)
 			continue
 		}
-		device.Protocols[CustomMetadata][trimKey] = trimValue // create or update a field in CustomMetadata
+
+		if _, found := device.Protocols[CustomMetadata]; !found {
+			metadata := make(contract.ProtocolProperties)
+			device.Protocols[CustomMetadata] = metadata
```
ajcasagrande: these can be simplified to 1 line

internal/driver/custommetadata.go
```diff
@@ -47,37 +52,38 @@ func (onvifClient *OnvifClient) getCustomMetadata(device contract.Device, data [
...
+		return device.Protocols[CustomMetadata], nil
 	}
 
+	// if a list of fields is provided, return those specific fields
+	metadataObj, err = onvifClient.getSpecificCustomMetadata(device, data)
+	if err != nil {
+		return nil, errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf("failed to get specific metadata for device %s", onvifClient.DeviceName), err)
+	}
 	return metadataObj, nil
+
```
ajcasagrande: remove blank line

internal/driver/custommetadata.go
```diff
@@ -47,37 +52,38 @@ func (onvifClient *OnvifClient) getCustomMetadata(device contract.Device, data [
...
 
-	err := json.Unmarshal(data, &dataArray)
+	err := json.Unmarshal(data, &input)
 	if err != nil {
 		return nil, errors.NewCommonEdgeX(errors.KindServerError, "failed to unmarshal the json request body", err)
 	}
-	if len(dataArray[CustomMetadata]) == 0 {
-		return nil, errors.NewCommonEdgeX(errors.KindServerError, "no data in query body", err)
+	if len(input[CustomMetadata]) == 0 {
+		return nil, errors.NewCommonEdgeX(errors.KindIOError, "no data in query body", err)
```
ajcasagrande: same as above, use `errors.KindContractInvalid`

internal/driver/custommetadata_test.go
```diff
@@ -34,244 +28,236 @@ func TestOnvifClient_getCustomMetadata(t *testing.T) {
...
+			false},
+		{
+			"happy path with data for single field",
+			myDevice,
+			`{"CustomMetadata":["CommonName"]}`,
+			contract.ProtocolProperties{
+				"CommonName": myDevice.Protocols[CustomMetadata]["CommonName"],
+			},
+			false,
+		},
```
ajcasagrande: all of these test cases should include the actual variable names, similar to the old code up above. 

example of what that would look like
```go
		{
			name: "happy path with data for single field",
			device: myDevice,
			data: `{"CustomMetadata":["CommonName"]}`,
			expected: contract.ProtocolProperties{
				"CommonName": myDevice.Protocols[CustomMetadata]["CommonName"],
			},
		},
```

internal/driver/custommetadata_test.go
```diff
@@ -34,244 +28,236 @@ func TestOnvifClient_getCustomMetadata(t *testing.T) {
...
+		{
+			"happy path with data for multiple field",
+			myDevice,
+			`{"CustomMetadata":["Location","CommonName"]}`,
+			contract.ProtocolProperties{
+				"Location":   myDevice.Protocols[CustomMetadata]["Location"],
+				"CommonName": myDevice.Protocols[CustomMetadata]["CommonName"],
+			},
+			false,
```
ajcasagrande: Instead of setting it to false using `errorExpected: false` for the happy paths, i would just leave that field out altogether (it is false by default anyways). 

This reduces the amount of code you have to write and makes it more noticeable when it is true, because your eyes will notice the `errorExpected: true` that wasn't there in the others.

internal/driver/custommetadata_test.go
```diff
@@ -34,244 +28,236 @@ func TestOnvifClient_getCustomMetadata(t *testing.T) {
...
-		{"happy path without data - getAll", myDevice, "", myDevice.Protocols[CustomMetadata], false},
-		{"happy path with data for single field", myDevice, singleDataTest.data, singleDataTest.expectedResponse, false},
-		{"happy path with data for multiple field", myDevice, happyMultipleDataTest.data, happyMultipleDataTest.expectedResponse, false},
-		{"happy path with data for non-existent field", myDevice, noFieldDataTest.data, noFieldDataTest.expectedResponse, false},
-		{"happy path with data for multiple non-existent fields", myDevice, noFieldsDataTest.data, noFieldsDataTest.expectedResponse, false},
-		{"with empty data", myDevice, emptyData.data, emptyData.expectedResponse, true},
-		{"badJson", myDevice, "bogus", myDevice.Protocols[CustomMetadata], true},
+		{
+			"happy path without data - getAll",
+			myDevice,
```
ajcasagrande: it looks like all of the test cases are using the same device. instead of passing it in to each function, I would recommend creating a generator function and calling it inside the actual test case (this also ensures you have no cross-contamination of tests.

I would also call it once at the beginning of the func so you can still use it as input to the expectedResponse.
```go
func getTestDevice() contract.Device {
return contract.Device{
		Protocols: map[string]contract.ProtocolProperties{
			CustomMetadata: {
				"Location":          "Front door",
				"CommonName":        "Front door camera",
				"Installation date": "01/01/2022",
				"Maintenance date":  "05/01/2022",
			},
		},
	}
}

File: internal/driver/custommetadata_test.go [hunk 26]
```diff
@@ -34,244 +28,236 @@ func TestOnvifClient_getCustomMetadata(t *testing.T) {
...
-				CustomMetadata: {},
+	tests := []struct {
+		name          string
+		device        contract.Device
+		data          string
+		expected      contract.ProtocolProperties
+		errorExpected bool
+	}{
+		{
+			"happy-path-withoutData",
```
ajcasagrande: all of the same comments from above apply here.

internal/driver/custommetadata_test.go
```diff
@@ -34,244 +28,236 @@ func TestOnvifClient_getCustomMetadata(t *testing.T) {
...
-			updatedDevice, err := onvifClient.setCustomMetadata(tt.test.device, []byte(tt.test.data))
-			if tt.errorExpected {
+			updatedDevice, err := onvifClient.setCustomMetadata(test.device, []byte(test.data))
+			if test.errorExpected {
 				require.Error(t, err)
 				return
 			}
 			require.NoError(t, err)
-			require.Equal(t, tt.test.expectedResponse, updatedDevice.Protocols[CustomMetadata])
+			require.Equal(t, test.expected, updatedDevice.Protocols[CustomMetadata])
```
ajcasagrande: change to `assert.Equal`. use `require` only when the test cannot continue without it passing. using `assert` lets you keep checking other things in case multiple errors are present.

ajcasagrande (changes requested):

internal/driver/onvifclient.go
```diff
@@ -137,12 +138,12 @@ func (onvifClient *OnvifClient) CallOnvifFunction(req sdkModel.CommandRequest, f
...
 		return nil, errors.NewCommonEdgeXWrapper(edgexErr)
 	}
+	custom, _ := attributeByKey(req.Attributes, "custom") // may be unneccessary
 	functionName, edgexErr := attributeByKey(req.Attributes, functionType)
 	if edgexErr != nil {
 		return nil, errors.NewCommonEdgeXWrapper(edgexErr)
 	}
-
-	if serviceName == EdgeXWebService {
+	if serviceName == EdgeXWebService || custom == "Custom" { //TODO: Make a constant
```
ajcasagrande: I recommend removing the `custom` and just using `serviceName == EdgeXWebService`

cmd/res/profiles/camera.yaml
```diff
@@ -162,6 +162,17 @@ deviceResources:
     properties:
       valueType: "Object"
       readWrite: "W"
+  - name: "CustomMetadata"
+    isHidden: false
+    description: "Get and set custom added information about the cameras"
+    attributes:
+      service: "Device"
+      custom: "Custom"
```
ajcasagrande: I recommend removing the custom field, and setting service to `EdgeX`. this will align better with the other custom funcs
```suggestion
      service: "EdgeX"
```

internal/driver/onvifclient.go
```diff
@@ -137,12 +138,12 @@ func (onvifClient *OnvifClient) CallOnvifFunction(req sdkModel.CommandRequest, f
 	if edgexErr != nil {
 		return nil, errors.NewCommonEdgeXWrapper(edgexErr)
 	}
+	custom, _ := attributeByKey(req.Attributes, "custom") // may be unneccessary
```
ajcasagrande: yeah we should remove

internal/driver/onvifclient.go
```diff
@@ -169,6 +170,37 @@ func (onvifClient *OnvifClient) CallOnvifFunction(req sdkModel.CommandRequest, f
...
+		device, err := sdk.RunningService().GetDeviceByName(deviceName)
+		if err != nil {
+			return nil, errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf("failed to get device '%s'", deviceName), err)
+		}
+		metadataObj, edgexError := onvifClient.getCustomMetadata(device, data)
+		if edgexError != nil {
+			onvifClient.driver.lc.Errorf("Failed to get custom metadata field from device %s", onvifClient.DeviceName)
+			return nil, edgexError
+		}
+		attributes[URLRawQuery] = "" // flush out the query so it resets with new calls
```
ajcasagrande: interesting. did you run into caching issues with this field? if so, maybe that speaks to a bug in some lower level or sdk code.

ajcasagrande (changes requested):
A few more comments after reviewing the final files.

cmd/res/profiles/camera.yaml
weichou1229: Yes, please refer to the doc and add the `GetCustomMetadata` and `SetCustomMetadata` to the README file.
https://github.com/edgexfoundry/device-onvif-camera/blob/0a79c5a718067479237195976121108915350925/README.md?plain=1#L97

internal/driver/onvifclient.go
ChristianDarr-personal: Yes, it seemed to cache the previous data, so if I ran a command with query data and then without, it would not work properly.

internal/driver/custommetadata.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil, errors.NewCommonEdgeX(errors.KindServerError, "failed to unmarshal the json request body", err)
+	}
+	if len(input) == 0 {
+		return nil, errors.NewCommonEdgeX(errors.KindContractInvalid, "no data in query body", err)
+	}
+
+	for _, key := range input {
+		value, found := device.Protocols[CustomMetadata][key]
+		if !found {
+			onvifClient.driver.lc.Warnf("Failed to find custom metadata field %s", key) // TODO: should this also be displayed in command response?
```
brianointel: should this "TODO" still be here?

brianointel (dismissed):

internal/driver/custommetadata.go
```diff
@@ -15,68 +15,77 @@ import (
...
 	if err != nil {
 		return device, errors.NewCommonEdgeX(errors.KindServerError, "failed to unmarshal the json request body", err)
 	}
 	if len(dataObj) == 0 {
-		return device, errors.NewCommonEdgeX(errors.KindIOError, "no data in PUT command", err)
+		return device, errors.NewCommonEdgeX(errors.KindContractInvalid, "no data in PUT command", err)
+	}
+
+	for okey, ovalue := range device.Protocols[OnvifProtocol] {
+		fmt.Printf("%s, %s", okey, ovalue)
```
ajcasagrande: was this just for testing? It should be removed or changed to Debug log level.

internal/driver/custommetadata.go
```diff
@@ -15,68 +15,77 @@ import (
...
 	response := make(map[string]string)
 
 	err := json.Unmarshal(data, &input)
 	if err != nil {
 		return nil, errors.NewCommonEdgeX(errors.KindServerError, "failed to unmarshal the json request body", err)
 	}
-	if len(input[CustomMetadata]) == 0 {
-		return nil, errors.NewCommonEdgeX(errors.KindIOError, "no data in query body", err)
+	if len(input) == 0 {
+		return nil, errors.NewCommonEdgeX(errors.KindContractInvalid, "no data in query body", err)
```
ajcasagrande:
```suggestion
		return nil, errors.NewCommonEdgeX(errors.KindContractInvalid, "no data in query parameter", err)
```

internal/driver/custommetadata.go
```diff
@@ -87,3 +96,29 @@ func (onvifClient *OnvifClient) getSpecificCustomMetadata(device contract.Device
...
+func (onvifClient *OnvifClient) deleteCustomMetadata(device contract.Device, data []byte) (contract.Device, errors.EdgeX) {
+	device = onvifClient.initCustomMetadata(device)
+
+	var input []string
+	err := json.Unmarshal(data, &input)
+	if err != nil {
+		return device, errors.NewCommonEdgeX(errors.KindServerError, "failed to unmarshal the json request body", err)
+	}
+	if len(input) == 0 {
+		return device, errors.NewCommonEdgeX(errors.KindContractInvalid, "no data in query body", err)
```
ajcasagrande: its not a query body. should be named something else

internal/driver/custommetadata.go
```diff
@@ -87,3 +96,29 @@ func (onvifClient *OnvifClient) getSpecificCustomMetadata(device contract.Device
...
+	}
+
+	for _, deleteKey := range input {
+		deleteKey = strings.TrimSpace(deleteKey)
+		_, found := device.Protocols[CustomMetadata][deleteKey]
+		if !found {
```
ajcasagrande: i would combine these to one line

internal/driver/onvifclient.go
```diff
@@ -176,31 +175,50 @@ func (onvifClient *OnvifClient) callCustomFunction(resourceName, serviceName, fu
...
 			onvifClient.driver.lc.Errorf("Failed to get custom metadata field from device %s", onvifClient.DeviceName)
 			return nil, edgexError
 		}
-		attributes[URLRawQuery] = "" // flush out the query so it resets with new calls
 		cv, err = sdkModel.NewCommandValue(resourceName, common.ValueTypeObject, metadataObj)
 		if err != nil {
 			return nil, errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf("failed to create commandValue for the web service '%s' function '%s'", serviceName, functionName), err)
 		}
+
+		attributes[URLRawQuery] = "" // flush out the query so it resets with new calls
```
ajcasagrande, line 189: Im actually wondering if this is something that needs to be done for all of them? i don't see why it would only affect 1. oh... wait maybe because this is the only one that supports both having and not having a query param.

I think maybe we need to file a bug with EdgeX

internal/driver/custommetadata.go
```diff
@@ -0,0 +1,124 @@
...
+func (onvifClient *OnvifClient) initCustomMetadata(device contract.Device) contract.Device {
+	if _, found := device.Protocols[CustomMetadata]; !found {
+		saveOnvifProtocol := device.Protocols[OnvifProtocol]
+		device.Protocols = map[string]contract.ProtocolProperties{
+			OnvifProtocol:  saveOnvifProtocol,
+			CustomMetadata: {},
+		}
+	}
+	return device
+}
```
ajcasagrande: Can be simplified, and also I think may make more sense to be a pointer:

```suggestion
func (onvifClient *OnvifClient) initCustomMetadata(device *contract.Device) {
	if _, found := device.Protocols[CustomMetadata]; !found {
		device.Protocols[CustomMetadata] = contract.ProtocolProperties{}
	}
}
```

internal/driver/custommetadata.go
```diff
@@ -0,0 +1,124 @@
...
+
+// setCustomMetadata will return a map containing the fields provided in the call to the function
+func (onvifClient *OnvifClient) setCustomMetadata(device contract.Device, data []byte) (contract.Device, errors.EdgeX) {
+	device = onvifClient.initCustomMetadata(device)
```
ajcasagrande: if you make the `initCustomMetadata` take a pointer, change this (and all other usages) to 
```suggestion
	onvifClient.initCustomMetadata(&device)
```

ajcasagrande (changes requested):
Looks pretty good. Some thoughts.

internal/driver/onvifclient.go
ChristianDarr-personal: That was my assumption too, I tried replicating it with other commands but only because this one supports both does the issue come up.

internal/driver/custommetadata.go
ChristianDarr-personal: I can't assign anything to device.Protocols[CustomMetadata] in the case that CustomMetadata doesn't exist, because it will cause a panic. It's almost guaranteed to exist at this point unless the user has predefined devices that don't include CustomMetadata, as this is already initialized in discovery. That's why I was resetting the entire device.Protocols. Do you see an alternative to what I had in that case?

ajcasagrande: You can't set a value on the `CustomMetadata`, but you can set `CustomMetadata` itself. See the difference:

```go
device.Protocols[CustomMetadata]["Location"] = "Exit"   <== BAD

device.Protocols[CustomMetadata] = contract.ProtocolProperties{}   <== OK

device.Protocols[CustomMetadata] = contract.ProtocolProperties{
    "Location": "Exit",
}   <== OK

```

ChristianDarr-personal: I see my issue now, the panic I was seeing was due to something in my test setup. Thank you!

internal/driver/custommetadata.go
```diff
@@ -38,16 +34,15 @@ func (onvifClient *OnvifClient) setCustomMetadata(device contract.Device, data [
...
 	for key, value := range dataObj {
 		value = strings.TrimSpace(value)
 		key = strings.TrimSpace(key)
 		if len(key) == 0 {
-			onvifClient.driver.lc.Warn("tried to add an empty key: {\"\": \"%s\"}", value)
-			continue
+			inputErr := error.New("tried to add an empty key")
+			return saveDevice, errors.NewCommonEdgeX(errors.KindContractInvalid, "tried to add an empty key to Custom Metadata", inputErr)
+			// onvifClient.driver.lc.Warn("tried to add an empty key: {\"\": \"%s\"}", value)
+			// continue
```
ajcasagrande: remove commented out code.

internal/driver/custommetadata.go
```diff
@@ -99,21 +94,20 @@ func (onvifClient *OnvifClient) getSpecificCustomMetadata(device contract.Device
...
+	onvifClient.initCustomMetadata(&device)
 
 	var input []string
 	err := json.Unmarshal(data, &input)
 	if err != nil {
 		return device, errors.NewCommonEdgeX(errors.KindServerError, "failed to unmarshal the json request body", err)
 	}
 	if len(input) == 0 {
-		return device, errors.NewCommonEdgeX(errors.KindContractInvalid, "no data in query body", err)
+		return device, errors.NewCommonEdgeX(errors.KindContractInvalid, "no data in query parameter", err)
```
ajcasagrande: I believe the Delete call is a `PUT` right? then it should be `request body`.

```suggestion
		return device, errors.NewCommonEdgeX(errors.KindContractInvalid, "no data in request body", err)
```

ajcasagrande (approved):
LGTM

PR #85, opened 2022-06-17
build: Upgrade sdk and paho.mqtt lib

weichou1229:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features) **not impact the test**
- [ ] Docs have been added / updated (for bug fixes / features) **not impact the doc**

**If your build fails** due to your commit message not passing the build checks, please review the guidelines here: https://github.com/edgexfoundry/edgex-examples/blob/master/.github/CONTRIBUTING.md.

## What is the new behavior?

- Upgrade sdk to v2.2.1-dev.5
- Upgrade paho.mqtt to v1.4.1

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## New Imports
Are there any new imports or modules? If so, what are they used for and why?

## Specific Instructions
Are there any specific instructions or things that should be known prior to reviewing?

## Other information

PR #89, opened 2022-06-23
Revert "build: Upgrade sdk and paho.mqtt lib"

cloudxxx8:
Reverts edgexfoundry/device-onvif-camera#85

Revert the changes because this will be released with v2.2, and we will update it again after release


weichou1229 (approved):
LGTM

PR #90, opened 2022-06-23
feat: mac credential storage

brianointel:
## PR Checklist

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

- [ ] Yes
- [ ] No

## New Imports

## Specific Instructions

## Other information

### DIFF
--- a/internal/driver/checkstatuses.go
+++ b/internal/driver/checkstatuses.go
@@ -12,7 +12,7 @@ import (
 
 	"github.com/IOTechSystems/onvif"
 	"github.com/edgexfoundry/device-sdk-go/v2/pkg/service"
-	sdkModel "github.com/edgexfoundry/go-mod-core-contracts/v2/models"
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/models"
 )
 
 // checkStatuses loops through all registered devices and tries to determine the most accurate connection state
@@ -25,8 +25,24 @@ func (d *Driver) checkStatuses() {
 
 		status := d.testConnectionMethods(device)
 
-		if err := d.updateDeviceStatus(device.Name, status); err != nil {
+		if statusChanged, err := d.updateDeviceStatus(device.Name, status); err != nil {
 			d.lc.Warnf("Could not update device status for device %s: %s", device.Name, err.Error())
+		} else if statusChanged && status == UpWithAuth {
+			d.lc.Infof("Device %s is now %s, refreshing the device information.", device.Name, UpWithAuth)
+			go func(device models.Device) {
+				// todo: should we update the device?
+				err2 := d.refreshDeviceInformation(device)
+				if err2 != nil {
+					d.lc.Warnf("An error occurred while refreshing the device information for %s: %s",
+						device.Name, err2.Error())
+				}
+
+				err2 = d.refreshNetworkInterfaces(device)
+				if err2 != nil {
+					d.lc.Warnf("An error occurred while refreshing the network information for %s: %s",
+						device.Name, err2.Error())
+				}
+			}(device)
 		}
 	}
 }
@@ -35,7 +51,7 @@ func (d *Driver) checkStatuses() {
 // and return the most accurate status
 // Higher degrees of connection are tested first, becuase if they
 // succeed, the lower levels of connection will too
-func (d *Driver) testConnectionMethods(device sdkModel.Device) (status string) {
+func (d *Driver) testConnectionMethods(device models.Device) (status string) {
 
 	// sends get capabilities command to device (does not require credentials)
 	devClient, edgexErr := d.newTemporaryOnvifClient(device)
@@ -61,7 +77,7 @@ func (d *Driver) testConnectionMethods(device sdkModel.Device) (status string) {
 
 // tcpProbe attempts to make a connection to a specific ip and port list to determine
 // if there is a service listening at that ip+port.
-func (d *Driver) tcpProbe(device sdkModel.Device) bool {
+func (d *Driver) tcpProbe(device models.Device) bool {
 	proto, ok := device.Protocols[OnvifProtocol]
 	if !ok {
 		d.lc.Warnf("Device %s is missing required %s protocol info, cannot send probe.", device.Name, OnvifProtocol)
@@ -85,7 +101,7 @@ func (d *Driver) tcpProbe(device sdkModel.Device) bool {
 	return true
 }
 
-func (d *Driver) updateDeviceStatus(deviceName string, status string) error {
+func (d *Driver) updateDeviceStatus(deviceName string, status string) (bool, error) {
 	// todo: maybe have connection levels known as ints, so that way we can log at different levels based on
 	// if the connection level went up or down
 	shouldUpdate := false
@@ -95,14 +111,16 @@ func (d *Driver) updateDeviceStatus(deviceName string, status string) error {
 	if err != nil {
 		d.lc.Errorf("Unable to get device %s from cache while trying to update its status to %s. Error: %s",
 			device.Name, status, err.Error())
-		return err
+		return false, err
 	}
 
+	statusChanged := false
 	oldStatus := device.Protocols[OnvifProtocol][DeviceStatus]
 	if oldStatus != status {
 		d.lc.Infof("Device status for %s is now %s (used to be %s)", device.Name, status, oldStatus)
 		device.Protocols[OnvifProtocol][DeviceStatus] = status
 		shouldUpdate = true
+		statusChanged = true
 	}
 
 	if status != Unreachable {
@@ -111,10 +129,10 @@ func (d *Driver) updateDeviceStatus(deviceName string, status string) error {
 	}
 
 	if shouldUpdate {
-		return service.RunningService().UpdateDevice(device)
+		return statusChanged, service.RunningService().UpdateDevice(device)
 	}
 
-	return nil
+	return statusChanged, nil
 }
 
 // taskLoop manages all of our custom background tasks such as checking camera statuses at regular intervals
--- a/internal/driver/checkstatuses.go
+++ b/internal/driver/checkstatuses.go
@@ -100,6 +100,7 @@ func (d *Driver) tcpProbe(device models.Device) bool {
 	return true
 }
 
+// updateDeviceStatus updates the status of a device in the cache. Returns true if the status changed. Returns any errors that occur if failure.
 func (d *Driver) updateDeviceStatus(deviceName string, status string) (bool, error) {
 	// todo: maybe have connection levels known as ints, so that way we can log at different levels based on
 	// if the connection level went up or down
--- a/internal/driver/checkstatuses.go
+++ b/internal/driver/checkstatuses.go
@@ -29,26 +29,28 @@ func (d *Driver) checkStatuses() {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
+
 			status := d.testConnectionMethods(device)
+			if statusChanged, updateDeviceStatusErr := d.updateDeviceStatus(device.Name, status); updateDeviceStatusErr != nil {
+				d.lc.Warnf("Could not update device status for device %s: %s", device.Name, updateDeviceStatusErr.Error())
 
-		if statusChanged, updateDeviceStatusErr := d.updateDeviceStatus(device.Name, status); updateDeviceStatusErr != nil {
-			d.lc.Warnf("Could not update device status for device %s: %s", device.Name, updateDeviceStatusErr.Error())
-		} else if statusChanged && status == UpWithAuth {
-			d.lc.Infof("Device %s is now %s, refreshing the device information.", device.Name, UpWithAuth)
-			go func(device models.Device) {
-				refreshErr := d.refreshDeviceInformation(device)
-				if refreshErr != nil {
-					d.lc.Warnf("An error occurred while refreshing the device information for %s: %s",
-						device.Name, refreshErr.Error())
-				}
+			} else if statusChanged && status == UpWithAuth {
+				d.lc.Infof("Device %s is now %s, refreshing the device information.", device.Name, UpWithAuth)
+				go func() { // refresh the device information in the background
+					refreshErr := d.refreshDeviceInformation(device)
+					if refreshErr != nil {
+						d.lc.Warnf("An error occurred while refreshing the device information for %s: %s",
+							device.Name, refreshErr.Error())
+					}
 
-				refreshErr = d.refreshNetworkInterfaces(device)
-				if refreshErr != nil {
-					d.lc.Warnf("An error occurred while refreshing the network information for %s: %s",
-						device.Name, refreshErr.Error())
-				}
-			}(device)
-		}
+					refreshErr = d.refreshNetworkInterfaces(device)
+					if refreshErr != nil {
+						d.lc.Warnf("An error occurred while refreshing the network information for %s: %s",
+							device.Name, refreshErr.Error())
+					}
+				}()
+			}
+		}()
 	}
 	wg.Wait()
 }
--- a/internal/driver/checkstatuses.go
+++ b/internal/driver/checkstatuses.go
@@ -30,7 +30,6 @@ func (d *Driver) checkStatuses() {
 		} else if statusChanged && status == UpWithAuth {
 			d.lc.Infof("Device %s is now %s, refreshing the device information.", device.Name, UpWithAuth)
 			go func(device models.Device) {
-				// todo: should we update the device?
 				err2 := d.refreshDeviceInformation(device)
 				if err2 != nil {
 					d.lc.Warnf("An error occurred while refreshing the device information for %s: %s",
--- a/internal/driver/checkstatuses.go
+++ b/internal/driver/checkstatuses.go
@@ -25,21 +25,21 @@ func (d *Driver) checkStatuses() {
 
 		status := d.testConnectionMethods(device)
 
-		if statusChanged, err := d.updateDeviceStatus(device.Name, status); err != nil {
-			d.lc.Warnf("Could not update device status for device %s: %s", device.Name, err.Error())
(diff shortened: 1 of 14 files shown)


internal/driver/checkstatuses.go
```diff
@@ -25,8 +25,24 @@ func (d *Driver) checkStatuses() {
 
 		status := d.testConnectionMethods(device)
 
-		if err := d.updateDeviceStatus(device.Name, status); err != nil {
+		if statusChanged, err := d.updateDeviceStatus(device.Name, status); err != nil {
 			d.lc.Warnf("Could not update device status for device %s: %s", device.Name, err.Error())
+		} else if statusChanged && status == UpWithAuth {
+			d.lc.Infof("Device %s is now %s, refreshing the device information.", device.Name, UpWithAuth)
+			go func(device models.Device) {
+				// todo: should we update the device?
```
lenny-goodell: Yes, looks like the Device data changed,  so Core Metadata needs that updated data.
```
device.Protocols[OnvifProtocol][DeviceStatus] = status
```

internal/driver/credentials.go
```diff
@@ -0,0 +1,118 @@
...
+
+	return credentials, err
+}
+
+func (d *Driver) getCredentialsFromMac(mac string) (Credentials, errors.EdgeX) {
+	if mac == "" {
+		credential, edgexErr := tryGetCredentials(d.config.AppCustom.DefaultSecretPath)
+		if edgexErr != nil {
+			d.lc.Error("failed to get credentials from default secret path", "err", edgexErr)
+			return Credentials{}, errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf("failed to get credentials from mac address %s", mac), edgexErr)
```
lenny-goodell:
```suggestion
			return Credentials{}, errors.NewCommonEdgeX(errors.KindServerError, "failed to get default credentials for empty mac address), edgexErr)
```

internal/driver/credentials.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		d.lc.Debug("Using default credentials from default secret path")
```
lenny-goodell:
```suggestion
		d.lc.Debug("Using default credentials from default secret path for empty mac address")
```

internal/driver/driver.go
```diff
@@ -652,6 +616,23 @@ func addressAndPort(xaddr string) (string, string) {
 	}
 }
 
+// todo: this should be integrated better with getDeviceInformation to avoid creating another temporary client
```
lenny-goodell, line 619: Yes! Why was this not done now?

internal/driver/macmapper_test.go
```diff
@@ -0,0 +1,57 @@
...
+	for _, test := range tests {
+		test := test
+		t.Run(test.mac, func(t *testing.T) {
+			sanitized, err := SanitizeMACAddress(test.mac)
+			assert.Equal(t, test.sanitized, sanitized)
+			if test.err {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
```
lenny-goodell:
```suggestion
			if test.err {
				require.Error(t, err)
                                return
			require.NoError(t, err)
			assert.Equal(t, test.sanitized, sanitized)
```

lenny-goodell (dismissed):

internal/driver/credentials.go
```diff
@@ -0,0 +1,118 @@
...
+			err.Error(),
+			timer.RemainingAsString())
+		timer.SleepForInterval()
+	}
+
+	return credentials, err
+}
+
+func (d *Driver) getCredentialsFromMac(mac string) (Credentials, errors.EdgeX) {
+	if mac == "" {
```
ajcasagrande: This code needs to be a bit refactored. The defaults should be returned if the mac is empty _OR_ if the mac does not exist in the lookup table.

While refactoring, we may want to inline the `TryGetCredentialsForMACAddress` and make it part of this.

internal/driver/driver.go
ajcasagrande: Mostly because it was quick and dirty (which isn't a good excuse :wink: ), but also because that whole client and temporaryOnvifClient concept needs a good refactoring.

internal/driver/driver.go
```diff
@@ -712,3 +690,45 @@ func (d *Driver) newTemporaryOnvifClient(device models.Device) (*OnvifClient, er
 	}
 	return client, nil
 }
+
+// refreshNetworkInterfaces will attempt to retrieve the device information for the specified camera
+// and update the values in the protocol properties
```
ajcasagrande:
```suggestion
// refreshNetworkInterfaces will attempt to retrieve the mac address for the specified camera
// and update its value in the protocol properties
```

internal/driver/driver.go
```diff
@@ -712,3 +690,45 @@ func (d *Driver) newTemporaryOnvifClient(device models.Device) (*OnvifClient, er
...
+
+	// update device to latest version in cache to prevent race conditions
+	device, edgeXErr := sdk.RunningService().GetDeviceByName(device.Name)
+	if err != nil {
+		return edgeXErr
+	}
+
+	device.Protocols[OnvifProtocol][MACAddress] = string(netInfo.NetworkInterfaces.Info.HwAddress)
+
+	return sdk.RunningService().UpdateDevice(device)
```
ajcasagrande: we should check if it changed and only update the device if it has.

internal/driver/driver.go
```diff
@@ -712,3 +690,45 @@ func (d *Driver) newTemporaryOnvifClient(device models.Device) (*OnvifClient, er
...
+	device, edgeXErr := sdk.RunningService().GetDeviceByName(device.Name)
+	if err != nil {
+		return edgeXErr
+	}
+
+	device.Protocols[OnvifProtocol][Manufacturer] = devInfo.Manufacturer
+	device.Protocols[OnvifProtocol][Model] = devInfo.Model
+	device.Protocols[OnvifProtocol][FirmwareVersion] = devInfo.FirmwareVersion
+	device.Protocols[OnvifProtocol][SerialNumber] = devInfo.SerialNumber
+	device.Protocols[OnvifProtocol][HardwareId] = devInfo.HardwareId
```
ajcasagrande: same here, we should probably check if anything changed before sending a device update request.

internal/driver/macmapper.go
```diff
@@ -0,0 +1,114 @@
...
+// inverts it into a quick lookup map of mac address to secret path.
+func (m *MACAddressMapper) UpdateMappings(raw map[string]string) {
+	m.credsMu.Lock()
+	defer m.credsMu.Unlock()
+
+	lc := sdk.RunningService().LoggingClient
+
+	credsMap := make(map[string]string)
+	for secretPath, macs := range raw {
+		if _, err := tryGetCredentials(secretPath); err != nil {
```
ajcasagrande: This should probably just use the first part of `tryGetCredentials` which queries the `SecretProvider` directly. In the future it could make use of https://github.com/edgexfoundry/go-mod-bootstrap/issues/350 if implemented.

```suggestion
		if _, err := sdk.RunningService().SecretProvider.GetSecret(secretPath, UsernameKey, PasswordKey, AuthModeKey); err != nil {
```

internal/driver/macmapper.go
```diff
@@ -0,0 +1,114 @@
...
+			lc.Warnf("One or more MAC address mappings exist for the secret path '%s' which does not exist in the Secret Store!", secretPath)
+		}
+
+		for _, mac := range strings.Split(macs, ",") {
+			sanitized, err := SanitizeMACAddress(mac)
+			if err != nil {
+				lc.Warnf("Skipping entry: %s", err.Error())
```
ajcasagrande:
```suggestion
				lc.Warnf("Skipping invalid mac address %s: %s", mac, err.Error())
```

internal/driver/onvifdiscovery.go
```diff
@@ -133,6 +132,13 @@ func (d *Driver) createDiscoveredDevice(onvifDevice onvif.Device) (sdkModel.Disc
 		device.Protocols[OnvifProtocol][DeviceStatus] = UpWithAuth
 		device.Protocols[OnvifProtocol][LastSeen] = time.Now().Format(time.UnixDate)
 
+		netInfo, err := d.getNetworkInterfaces(device)
+		if err != nil {
+			d.lc.Warnf("failed to get the network information for the camera %s, %v", endpointRefAddr, edgexErr)
```
ajcasagrande: This should be moved below the deviceName name assignment, and the deviceName used in the error message instead of endpointRefAddr

internal/driver/macmapper.go
```diff
@@ -0,0 +1,114 @@
...
+	i := 0
+	for mac := range m.credsMap {
+		macs[i] = mac
+		i++
+	}
+
+	return macs
+}
+
+func (m *MACAddressMapper) GetSecretPathForMACAddress(mac string) (string, error) {
```
ajcasagrande: we should comment all exported functions

internal/driver/credentials.go
```diff
@@ -0,0 +1,118 @@
...
+			secretPath,
+			err.Error(),
+			timer.RemainingAsString())
+		timer.SleepForInterval()
+	}
+
+	return credentials, err
+}
+
+func (d *Driver) getCredentialsFromMac(mac string) (Credentials, errors.EdgeX) {
```
ajcasagrande: this should ideally be separated from the driver code in such a way that it can be unit tested. Not sure how much work that is though.

internal/driver/credentials.go
```diff
@@ -0,0 +1,118 @@
...
+
+	credentials := Credentials{
+		Username: secretData[UsernameKey],
+		Password: secretData[PasswordKey],
+		AuthMode: secretData[AuthModeKey],
+	}
+
+	if !IsAuthModeValid(secretData[AuthModeKey]) {
+		sdk.RunningService().LoggingClient.Warnf("AuthMode is set to an invalid value: %s. setting value to 'none'.", credentials.AuthMode)
+		credentials.AuthMode = AuthModeNone
```
ajcasagrande: Maybe we should default it to `usernametoken` or something sensible? If the creds are valid, but the mode isn't, I think we should still try and use them.

ajcasagrande (changes requested):

internal/driver/macmapper.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return Credentials{}, err
+	}
+	return tryGetCredentials(secretPath)
+}
+
+// SanitizeMACAddress takes in a MAC address in one of the IEEE 802 MAC-48, EUI-48, EUI-64 formats
+// and will return it in the standard go format, using colons and lower case letters:
+// Example:	aa:bb:cc:dd:ee:ff
+func SanitizeMACAddress(mac string) (string, error) {
```
ChristianDarr-personal, line 107: I would add white-space trimming to this, because if the macs are listed comma separated with spaces `("aa:bb:cc:dd:ee:ff, 11:22:33:44:55:66")`, this will throw an error I believe.

ChristianDarr-personal (commented):
Looks good, I just had one small comment.

internal/driver/driver.go
```diff
@@ -712,3 +690,59 @@ func (d *Driver) newTemporaryOnvifClient(device models.Device) (*OnvifClient, er
...
+	if err != nil {
+		return edgeXErr
+	}
+
+	if devInfo.Manufacturer != device.Protocols[OnvifProtocol][Manufacturer] ||
+		devInfo.Model != device.Protocols[OnvifProtocol][Model] ||
+		devInfo.FirmwareVersion != device.Protocols[OnvifProtocol][FirmwareVersion] ||
+		devInfo.SerialNumber != device.Protocols[OnvifProtocol][SerialNumber] ||
+		devInfo.HardwareId != device.Protocols[OnvifProtocol][HardwareId] {
+		device.Protocols[OnvifProtocol][Manufacturer] = devInfo.Manufacturer
```
ajcasagrande, line 737: the way go prefers to indent these makes it a bit confusing imo. I think a blank line after the `{` would make it more readable. and you can also get rid of the blank line before the updateDevice call.

internal/driver/macmapper_test.go
```diff
@@ -0,0 +1,56 @@
...
+			err: true,
+		},
+		{
+			mac:       "AA:BB:CC:DD:EE:FF",
+			sanitized: "aa:bb:cc:dd:ee:ff",
+		},
+		{
+			mac:       "11-22-33-44-55-66",
+			sanitized: "11:22:33:44:55:66",
+		},
```
ajcasagrande, line 45: add some test cases for the whitespace trimming

internal/driver/onvifclient.go
```diff
@@ -60,12 +60,9 @@ func (d *Driver) newOnvifClient(device models.Device) (*OnvifClient, errors.Edge
 		return nil, errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf("failed to create cameraInfo for camera %s", device.Name), edgexErr)
 	}
 
-	var credential config.Credentials
-	if cameraInfo.AuthMode != onvif.NoAuth {
-		credential, edgexErr = d.getCredentials(cameraInfo.SecretPath)
-		if edgexErr != nil {
-			return nil, errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf("failed to get credentials for camera %s", device.Name), edgexErr)
-		}
+	credential, edgexErr := d.tryGetCredentialsFromMac(device.Protocols[OnvifProtocol][MACAddress])
```
ajcasagrande: I think this should be removed for now and put in a new branch for next sprint.

internal/driver/onvifdiscovery.go
```diff
@@ -133,12 +132,20 @@ func (d *Driver) createDiscoveredDevice(onvifDevice onvif.Device) (sdkModel.Disc
 		device.Protocols[OnvifProtocol][DeviceStatus] = UpWithAuth
 		device.Protocols[OnvifProtocol][LastSeen] = time.Now().Format(time.UnixDate)
 
+		netInfo, err := d.getNetworkInterfaces(device)
```
ajcasagrande: this should be moved down lower by where the err is actually checked.

internal/driver/checkstatuses.go
ajcasagrande: im not sure what this todo is for. maybe it needs to be removed?

internal/driver/credentials.go
```diff
@@ -0,0 +1,125 @@
...
+			timer.RemainingAsString())
+		timer.SleepForInterval()
+	}
+
+	return credentials, err
+}
+
+// tryGetCredentialsFromMac will attempt to retrieve the credentials associated with the given mac address.
+func (d *Driver) tryGetCredentialsFromMac(mac string) (Credentials, errors.EdgeX) {
+	if mac == "" {
```
ajcasagrande: this still isn't what I had in mind, as there is a lot of duplicated code. this is also why i meantioned inlining the `TryGetCredentialsForMACAddress` as maybe you can just use the different flows to define the `secretPath`, and then they all flow into one block of code that does the lookup.

ajcasagrande (changes requested):

internal/driver/checkstatuses.go
ajcasagrande: Yeah I'm not so sure what that comment referred to. Each of the sub calls below, and the one above update core-metadata, so i don't think it is that.

I think it might have been a todo written before the code that follows it (ie. should we refresh the device information and network information upon a device coming back online), which has been done.

internal/driver/credentials.go
ajcasagrande: also, the mac needs to be sanitized first

ChristianDarr-personal (commented):
LGTM

internal/driver/checkstatuses.go
```diff
@@ -85,7 +100,7 @@ func (d *Driver) tcpProbe(device sdkModel.Device) bool {
 	return true
 }
 
-func (d *Driver) updateDeviceStatus(deviceName string, status string) error {
+func (d *Driver) updateDeviceStatus(deviceName string, status string) (bool, error) {
```
ajcasagrande, line 113: add a function comment to say that it returns whether or not the device status changed, and any errors that ocurred.

internal/driver/driver.go
ajcasagrande: Will address this todo as part of a future story.

internal/driver/macmapper.go
```diff
@@ -0,0 +1,113 @@
...
+				credsMap[sanitized] = secretPath
+			}
+		}
+	}
+
+	m.credsMap = credsMap
+}
+
+// ListMACAddresses will return a slice of mac addresses that have been assigned credentials
+func (m *MACAddressMapper) ListMACAddresses() []string {
```
ajcasagrande, line 65: Note that this was part of checking off the acceptence criteria, but the function isn't called, so not sure if we even need it.

internal/driver/checkstatuses.go
```diff
@@ -25,8 +25,23 @@ func (d *Driver) checkStatuses() {
 
 		status := d.testConnectionMethods(device)
 
-		if err := d.updateDeviceStatus(device.Name, status); err != nil {
+		if statusChanged, err := d.updateDeviceStatus(device.Name, status); err != nil {
 			d.lc.Warnf("Could not update device status for device %s: %s", device.Name, err.Error())
+		} else if statusChanged && status == UpWithAuth {
+			d.lc.Infof("Device %s is now %s, refreshing the device information.", device.Name, UpWithAuth)
+			go func(device models.Device) {
+				err2 := d.refreshDeviceInformation(device)
```
ChristianDarr-personal: Could this keep the name err? Or should the name be more descriptive if it does need to be named differently?

internal/driver/macmapper.go
brianointel: Todd clarified it will be needed for the next story 7236

ajcasagrande (dismissed):
LGTM

ajcasagrande (approved):
LGTM

PR #91, opened 2022-06-23
fix: run status checks for each device in parallel

### LINKED ISSUE #88: checkStatuses takes too long to complete
# 🐞 Bug Report

### Description and Minimal Reproduction [**REQUIRED**]
- Run `device-onvif-camera` and connect it to 10 simulated or real onvif cameras. 
- Disconnect the cameras
- Take note of how long it takes to complete the `checkStatuses` call

For me, this took about `44 seconds` when running as stated above.  Using multiple goroutines is able to shorten that to about 11 seconds, but still pretty long.

Obviously this will cause issues, as the default time period between checks are 30 seconds, so the code needs to complete before then.

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍  Your Environment
**Deployment Environment:**

**EdgeX Version [**REQUIRED**]:**

**Anything else relevant?**

ajcasagrande:
Signed-off-by: Anthony Casagrande <anthony.j.casagrande@intel.com>

Closes #88 

There are still optimizations that can be done, but with 10 dockerized simulators, the check completes in 7 seconds when they first go offline, and then 5 seconds everytime after that. This is opposed to the ~38 seconds they took before with the same setup.

### DIFF
--- a/internal/driver/checkstatuses.go
+++ b/internal/driver/checkstatuses.go
@@ -8,6 +8,7 @@ package driver
 
 import (
 	"net"
+	"sync"
 	"time"
 
 	"github.com/IOTechSystems/onvif"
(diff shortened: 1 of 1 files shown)


lenny-goodell (dismissed):
LGTM

ChristianDarr-personal (dismissed):
LGTM

internal/driver/checkstatuses.go
```diff
@@ -18,17 +19,24 @@ import (
...
 func (d *Driver) checkStatuses() {
 	d.lc.Debug("checkStatuses has been called")
+	wg := sync.WaitGroup{}
 	for _, device := range service.RunningService().Devices() {
 		if device.Name == d.serviceName { // skip control plane device
 			continue
 		}
 
-		status := d.testConnectionMethods(device)
+		device := device // save the device value within the closure
```
cloudxxx8: please use a different variable name for the async function to improve the readability and future maintenance

cloudxxx8 (changes requested):

internal/driver/checkstatuses.go
ajcasagrande: @cloudxxx8 I actually find that the re-use of the name makes it more maintainable, as it takes the outer `device` var out of scope, whereas with a different name, 2 variables are in scope and the developer could use the wrong one.

Having said that, I think I should move this line to first line of the for loop to prevent it from being lost/moved, etc.

cloudxxx8 (dismissed):
LGTM

cloudxxx8:
recheck

cloudxxx8:
@ajcasagrande the make test failed due to the formatting

[2022-06-28T08:54:32.099Z] gofmt -l $(find . -type f -name '*.go'| grep -v "/vendor/")
[2022-06-28T08:54:32.099Z] ./internal/driver/checkstatuses.go
[2022-06-28T08:54:32.099Z] [ "`gofmt -l $(find . -type f -name '*.go'| grep -v "/vendor/")`" = "" ]
[2022-06-28T08:54:32.099Z] make: *** [Makefile:57: test] Error 1

cloudxxx8 (approved):
LGTM

PR #149, opened 2022-08-29
test: unit test additions

brianointel:

## PR Checklist

- [X] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [ ] I am not introducing a new dependency (add notes below if you are)
- [ ] I have added unit tests for the new feature or bug fix (if not, why?)
- [ ] I have fully tested (add details below) this the new feature or bug fix (if not, why?)
- [ ] I have opened a PR for the related docs change (if not, why?)
  <link to docs PR>


internal/driver/config_test.go
```diff
@@ -51,3 +54,48 @@ func TestUpdateFromRaw(t *testing.T) {
...
+	tests := []struct {
+		input    map[string]models.ProtocolProperties
+		expected string
+
+		errorExpected bool
+	}{
+		{
+			input: map[string]models.ProtocolProperties{
+				OnvifProtocol: {
+					Address: "http://localhost",
```
ajcasagrande: typically the address does  not include the `http://` prefix and is an ip address.

internal/driver/config_test.go
```diff
@@ -51,3 +54,48 @@ func TestUpdateFromRaw(t *testing.T) {
 		})
 	}
 }
+
+// TestGetCameraXAddr verify the parsing of the camera XAddr
```
presatish:
```suggestion
// TestGetCameraXAddr verifies the parsing of the camera XAddr
```

internal/driver/config_test.go
```diff
@@ -51,3 +54,48 @@ func TestUpdateFromRaw(t *testing.T) {
...
+func TestGetCameraXAddr(t *testing.T) {
+
+	tests := []struct {
+		input    map[string]models.ProtocolProperties
+		expected string
+
```
presatish: extra line

brianointel:
@ajcasagrande: both @ChristianDarr-personal and I have a couple check-ins that fail the DCO check, should I still follow the rebase instructions?

internal/driver/config_test.go
```diff
@@ -51,3 +54,48 @@ func TestUpdateFromRaw(t *testing.T) {
...
+		errorExpected bool
+	}{
+		{
+			input: map[string]models.ProtocolProperties{
+				OnvifProtocol: {
+					Address: "http://localhost",
+					Port:    "80",
+				},
+			},
+			expected: "http://localhost:80",
```
presatish: I think its good to add `errorExpected` field to all inputs, it can be added here.

internal/driver/config_test.go
```diff
@@ -51,3 +54,48 @@ func TestUpdateFromRaw(t *testing.T) {
...
+				OnvifProtocol: {
+					CustomMetadata: "custommetadata",
+				},
+			},
+			errorExpected: true,
+		},
+		{
+			errorExpected: true,
+		},
+	}
```
presatish, line 104: One more test case that might be useful to test is passing empty string for address and checking for error.

presatish (changes requested):
Just started reviewing and have posted some comments.

internal/driver/config_test.go
brianointel: I think @ajcasagrande suggested we leave it off if it is `false`?

ajcasagrande: yes, I recommend leaving it off if it is `false`

internal/driver/config_test.go
```diff
@@ -51,3 +54,56 @@ func TestUpdateFromRaw(t *testing.T) {
...
+			input: map[string]models.ProtocolProperties{
+				OnvifProtocol: {
+					Address: "",
+					Port:    "",
+				},
+			},
+			errorExpected: true,
+		},
+		{
+			errorExpected: true,
```
ajcasagrande, line 102: this test is quite confusing. is the intent that the input is nil? if so, that is something I would prefer be explicit.

internal/driver/config_test.go
```diff
@@ -51,3 +54,56 @@ func TestUpdateFromRaw(t *testing.T) {
...
+		},
+	}
```
presatish, line 104: Also you can add a test case where the port is only empty and no error is expected.

internal/driver/credentials_test.go
```diff
@@ -7,20 +7,230 @@
...
+			}
+			require.NoError(t, err)
+			assert.Equal(t, test.expected.Username, actual.Username)
+			assert.Equal(t, test.expected.Password, actual.Password)
+			assert.Equal(t, test.expected.AuthMode, actual.AuthMode)
+		})
+	}
+}
+
+// TestTryGetCredentials verifies correct credentials are returned.
```
presatish: This comment needs to modified.

internal/driver/credentials_test.go
```diff
@@ -7,20 +7,230 @@
...
+	}{
+		{
+			path:     noAuthSecretPath,
+			username: "username",
+			password: "password",
+			authMode: onvif.DigestAuth,
+			expected: Credentials{
+				AuthMode: AuthModeNone,
+			},
+			errorExpected: false,
```
ajcasagrande: I would leave out the `false` errorExpected

internal/driver/subscriptionrequest_test.go
```diff
@@ -0,0 +1,45 @@
+package driver
```
presatish, line 7: License header missing.

internal/driver/credentials_test.go
```diff
@@ -7,20 +7,230 @@
...
+
+	mockSecretProvider := &mocks.SecretProvider{}
+
+	for i := range tests {
+		if tests[i].errorExpected {
+			mockSecretProvider.On("GetSecret", tests[i].path, UsernameKey, PasswordKey, AuthModeKey).Return(nil, errors.NewCommonEdgeX(errors.KindServerError, "unit test error", nil)).Once()
+
+		} else {
+			mockSecretProvider.On("GetSecret", tests[i].path, UsernameKey, PasswordKey, AuthModeKey).Return(map[string]string{"username": tests[i].username, "password": tests[i].password, "mode": tests[i].authMode}, nil).Once()
+		}
```
ajcasagrande: i feel like this makes more sense to call inside of the test run portion

internal/driver/driver_test.go
```diff
@@ -1,56 +1,265 @@
...
+// 			actual, err := d.HandleReadCommands(test.deviceName, test.protocols, test.reqs)
+// 			if test.errorExpected {
+// 				require.Error(t, err)
+// 			}
+// 			assert.Equal(t, test.expected, actual)
+// 		})
+// 	}
+// }
+
+// TestUpdateDevice: test for the proper updating of device information
```
presatish:
```suggestion
// TestUpdateDevice verifies proper updating of device information

```

internal/driver/onvifdiscovery_test.go
```diff
@@ -0,0 +1,349 @@
...
+			Name: "testDevice3", Protocols: map[string]models.ProtocolProperties{
+				OnvifProtocol: map[string]string{
+					EndpointRefAddress: "789",
+				},
+			},
+		},
+	}
+}
+
+func TestOnvifDiscovery_makeDeviceMap(t *testing.T) {
```
presatish, line 86: Remove `underscore` and use camel case naming to be consistent with other unit tests.

internal/driver/onvifdiscovery_test.go
```diff
@@ -0,0 +1,349 @@
...
+				Return("device-onvif-camera").Times(test.nameCalls)
+			devices := driver.makeDeviceRefMap()
+			mockService.AssertExpectations(t)
+
+			assert.Equal(t, devices, test.deviceMap)
+		})
+	}
+}
+
+func TestOnvifDiscovery_discoveryFilter(t *testing.T) {
```
presatish, line 192: Same comment as above for this too.

internal/driver/onvifdiscovery_test.go
```diff
@@ -0,0 +1,349 @@
...
+			mockService.On("Name").
+				Return("device-onvif-camera").Times(test.nameCalls)
+			devices := driver.makeDeviceRefMap()
+			mockService.AssertExpectations(t)
+
+			assert.Equal(t, devices, test.deviceMap)
+		})
+	}
+}
+
```
presatish, line 186: Add comment for this test case.

internal/driver/onvifdiscovery_test.go
```diff
@@ -0,0 +1,349 @@
...
+		{
+			Name: "testDevice3", Protocols: map[string]models.ProtocolProperties{
+				OnvifProtocol: map[string]string{
+					EndpointRefAddress: "789",
+				},
+			},
+		},
+	}
+}
+
```
presatish, line 76: Comment to briefly describe what this unit test achieves to be consistent with other unit tests.

internal/driver/credentials_test.go
```diff
@@ -7,20 +7,230 @@
...
+		expected      Credentials
+		errorExpected bool
+		username      string
+		password      string
+		authMode      string
+	}{
+		{
+			path:     noAuthSecretPath,
+			username: "username",
+			password: "password",
```
ajcasagrande: i feel like this should fail because the username password should be blank for noAuth

internal/driver/credentials_test.go
```diff
@@ -7,20 +7,230 @@
...
+				Password: "password",
+			},
+			errorExpected: false,
+		},
+		{
+			path:          "invalidPath",
+			username:      "username",
+			password:      "password",
+			authMode:      onvif.DigestAuth,
+			expected:      Credentials{},
```
ajcasagrande: you can leave out the expected on error cases where the expected is not being checked.

internal/driver/discoverymode_test.go
```diff
@@ -0,0 +1,115 @@
...
+package driver
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestIsValid verifies discovery mode setting.
+func TestIsValid(t *testing.T) {
```
ajcasagrande: TestDiscoveryMode_IsValid

internal/driver/discoverymode_test.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+}
+
+// TestIsMulticastEnabled verifies multicast setting.
+func TestIsMulticastEnabled(t *testing.T) {
+
+	tests := []struct {
+		input    DiscoveryMode
+		expected bool
+	}{
```
ajcasagrande: these two test functions can be combined if using something like this:
also, i recommend to rename `input` to `mode`
```suggestion
	tests := []struct {
		mode    DiscoveryMode
		multicast bool
		netscan bool
	}{
```

internal/driver/macmapper_test.go
```diff
@@ -178,3 +182,232 @@ func TestMatchEndpointRefAddressToMAC(t *testing.T) {
 		})
 	}
 }
+
```
presatish, line 185: Remove `Underscore` in name and add a comment to the test case above to be consistent with other unit tests.

internal/driver/macmapper_test.go
```diff
@@ -178,3 +182,232 @@ func TestMatchEndpointRefAddressToMAC(t *testing.T) {
...
+				ex2 := reflect.DeepEqual(test.alternateExpected, driver.macAddressMapper.credsMap)
+				assert.Equal(t, true, (ex1 || ex2))
+				return
+			}
+			assert.Equal(t, test.expected, driver.macAddressMapper.credsMap)
+		})
+	}
+}
+
+func TestMACAddressMapper_ListMACAddresses(t *testing.T) {
```
presatish: Same comment as above.

internal/driver/driver.go
```diff
@@ -1,825 +1,830 @@
-// -*- Mode: Go; indent-tabs-mode: t -*-
```
ajcasagrande: what happend here? Why does the whole file appear changed? line endings change?

internal/driver/driver_test.go
```diff
@@ -1,56 +1,265 @@
-// -*- Mode: Go; indent-tabs-mode: t -*-
```
ajcasagrande: same with this file, something messed up the change detection

internal/driver/driver_test.go
```diff
@@ -1,56 +1,265 @@
...
+		Attributes: map[string]interface{}{
+			URLRawQuery: fmt.Sprintf("%s=%s", jsonObject, base64EncodedStr),
+		},
+	}
+	data, err := parametersFromURLRawQuery(req)
+	require.NoError(t, err)
+	assert.Equal(t, parameters, string(data))
+}
+
+// TestAddressAndPort splits the address and port from a given string.
```
presatish:
```suggestion
// TestAddressAndPort verifies splitting of address and port from a given string.

```

internal/driver/subscriptionrequest_test.go
```diff
@@ -0,0 +1,45 @@
...
+		expected      time.Duration
+		errorExpected bool
+	}{
+		{
+			input:    "PT180S",
+			expected: 180000000000,
+		},
+		{
+			input:    "P1Y2M3W4DT5H6M7S",
+			expected: 18367000000000,
```
ajcasagrande, line 30: i would add more test cases that skips values in between since that is part of the support, like 1 year 5 hours. 5 days, 2 minutes. etc.

ajcasagrande (changes requested):
Still have more to go through. Comments so far.

internal/driver/macmapper_test.go
presatish, line 185: Remove Underscore in name and add a comment to the test case above to be consistent with other unit tests.

internal/driver/macmapper_test.go
```diff
@@ -178,3 +182,232 @@ func TestMatchEndpointRefAddressToMAC(t *testing.T) {
 		})
 	}
 }
+
+func TestMACAddressMapper_UpdateMappings(t *testing.T) {
```
presatish, line 186: Remove Underscore in name and add a comment to the test case above to be consistent with other unit tests.

internal/driver/driver_test.go
```diff
@@ -1,56 +1,265 @@
...
+			if test.errorExpected {
+				require.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+		})
+	}
+}
+
+func TestDriver_RemoveDevice(t *testing.T) {
```
presatish: Add comment and fix the naming by removing `underscore` to be consistent with other unit tests.

internal/driver/driver_test.go
```diff
@@ -1,56 +1,271 @@
...
+// 			actual, err := d.HandleReadCommands(test.deviceName, test.protocols, test.reqs)
+// 			if test.errorExpected {
+// 				require.Error(t, err)
+// 			}
+// 			assert.Equal(t, test.expected, actual)
+// 		})
+// 	}
+// }
+
+// TestUpdateDevice: verifies proper updating of device information
```
presatish:
```suggestion
// TestUpdateDevice verifies proper updating of device information

```

internal/driver/macmapper_test.go
ajcasagrande: actually I think this is a common naming scheme, GoLand recommends these names itself. `Test<TypeName>_<FuncName>`

internal/driver/driver_test.go
ajcasagrande: like the other func, I think the name is fine.

internal/driver/config_test.go
```diff
@@ -11,7 +14,7 @@ func TestUpdateFromRaw(t *testing.T) {
 			RequestTimeout:             5,
 			DefaultSecretPath:          "default_secret",
 			DiscoveryEthernetInterface: "eth0",
-			BaseNotificationURL:        "http://localhost:59984",
+			BaseNotificationURL:        "localhost:59984",
```
ajcasagrande: revert this change.

internal/driver/credentials_test.go
```diff
@@ -7,20 +7,222 @@
...
+
+	mockSecretProvider := &mocks.SecretProvider{}
+	mockService.On("GetSecretProvider").Return(mockSecretProvider)
+
+	for _, test := range tests {
+		test := test
+		t.Run(test.path, func(t *testing.T) {
+			if test.errorExpected {
+				mockSecretProvider.On("GetSecret", test.path, UsernameKey, PasswordKey, AuthModeKey).Return(nil, errors.NewCommonEdgeX(errors.KindServerError, "unit test error", nil)).Once()
+
```
ajcasagrande: remove blank line

internal/driver/driver_test.go
```diff
@@ -1,56 +1,271 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	testDeviceName = "test-device"
+)
+
+func createDriverWithMockService() (*Driver, *mocks.DeviceServiceSDK) {
+
```
ajcasagrande: remove blank line

internal/driver/driver_test.go
```diff
@@ -1,56 +1,271 @@
...
+		t.Run(test.device.Name, func(t *testing.T) {
+
+			if test.removeDeviceFailExpected {
+				mockService.On("RemoveDeviceByName", test.device.Name).Return("error").Once()
+			} else {
+				mockService.On("RemoveDeviceByName", test.device.Name).Return(nil).Once()
+			}
+			mockService.On("RemoveDeviceByName", test.device.Name).Return(nil).Once()
+			mockService.On("AddDevice", test.expectedDevice).Return(test.expectedDevice.Name, nil).Once()
+			mockService.On("UpdateDevice", test.device).Return(nil).Once()
```
ajcasagrande: i think these should be either-or, and it should check whether the correct ones were called using the mock assert expectations.

internal/driver/driver_test.go
```diff
@@ -1,56 +1,271 @@
...
+			},
+			devInfo: &device.GetDeviceInformationResponse{
+				Manufacturer:    "Intel",
+				Model:           "SimCamera",
+				FirmwareVersion: "2.5a",
+				SerialNumber:    "9a32410c",
+				HardwareId:      "1.0",
+			},
+			expectedDevice: contract.Device{
+				Name: "Intel-SimCamera-",
```
ajcasagrande: we should add an endpoint reference to have it be appended to the name, right?

internal/driver/driver_test.go
ajcasagrande: this func needs a comment specifically that it is testing the difference between removing a regular device and the control plane device.

internal/driver/driver_test.go
```diff
@@ -1,56 +1,271 @@
...
+	tests := []struct {
+		name       string
+		deviceName string
+		protocols  map[string]models.ProtocolProperties
+		wantErr    bool
+	}{
+		{
+			name:       "control plane device",
+			deviceName: "device-onvif-camera",
+			protocols:  map[string]models.ProtocolProperties{},
```
ajcasagrande: since the protocols field is ignored, remove it from the input table and hard-code it into the call below

internal/driver/macmapper.go
```diff
@@ -122,10 +121,6 @@ func macAddressBytewiseReverse(mac string) (string, error) {
 		return "", err
 	}
 	mac = strings.ReplaceAll(mac, ":", "")
-	if len(mac)%2 != 0 {
-		return "", fmt.Errorf("mac address %s has invalid length of %d", mac, len(mac))
-	}
-
 	buf := strings.Builder{}
 	// loop through the string backwards two characters at a time (1-byte)
```
ajcasagrande:
```suggestion
	// loop through the string backwards two characters at a time (1-byte). Since the MAC address is
	// already sanitized, we are guaranteed to have an even number of characters.
```

internal/driver/macmapper_test.go
```diff
@@ -178,3 +182,233 @@ func TestMatchEndpointRefAddressToMAC(t *testing.T) {
...
+			expected: map[string]string{
+				"aa:bb:cc:dd:ee:ff": "creds1",
+				"11:22:33:44:55:66": "creds2",
+				"ff:ee:dd:cc:bb:aa": "creds3",
+			},
+		},
+		{
+			name: "Add invalid macs",
+			currentMap: map[string]string{
+				"creds3": "FF:EE:DD:CC:BB:AA,asbc,asdf",
```
ajcasagrande: you are missing unit tests where there are multiple valid mac addresses in the same group.

internal/driver/credentials_test.go
presatish: Its passing with these current values and even blank ones.

internal/driver/macmapper_test.go
```diff
@@ -178,3 +182,233 @@ func TestMatchEndpointRefAddressToMAC(t *testing.T) {
...
+				"creds1": "AA:BB:CC:DD:EE:FF",
+				"creds2": "11:22:33:44:55:66",
+			},
+			expected: map[string]string{
+				"aa:bb:cc:dd:ee:ff": "creds1",
+				"11:22:33:44:55:66": "creds2",
+				"ff:ee:dd:cc:bb:aa": "creds3",
+			},
```
ajcasagrande: inst this the same as the above test?

internal/driver/macmapper_test.go
```diff
@@ -178,3 +182,233 @@ func TestMatchEndpointRefAddressToMAC(t *testing.T) {
...
+
+			driver, mockService := createDriverWithMockService()
+			driver.macAddressMapper = NewMACAddressMapper(mockService)
+			driver.macAddressMapper.credsMap = test.currentMap
+			mockSecretProvider := &mocks.SecretProvider{}
+			mockLoggingClient := logger.MockLogger{}
+
+			if len(test.currentMap) == 0 {
+				return
+			}
```
ajcasagrande: this should be a `require` because if it hits here that means we have an invalid test.

internal/driver/macmapper_test.go
```diff
@@ -178,3 +182,233 @@ func TestMatchEndpointRefAddressToMAC(t *testing.T) {
...
+	}
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+
+			driver, mockService := createDriverWithMockService()
+			driver.macAddressMapper = NewMACAddressMapper(mockService)
+			driver.macAddressMapper.credsMap = test.currentMap
+			mockSecretProvider := &mocks.SecretProvider{}
+			mockLoggingClient := logger.MockLogger{}
```
ajcasagrande:
```suggestion
			mockLoggingClient := logger.NewMockClient()
```

internal/driver/driver_test.go
ChristianDarr-personal: https://github.com/edgexfoundry/device-onvif-camera/pull/149/commits/9d5503b2ac70f45a7f1029cf620ebc15ff256cac

This the commit

internal/driver/macmapper_test.go
```diff
@@ -178,3 +182,233 @@ func TestMatchEndpointRefAddressToMAC(t *testing.T) {
...
+						Return(nil, nil)
+				}
+			}
+
+			mockService.On("GetSecretProvider").
+				Return(mockSecretProvider)
+			mockService.On("GetLoggingClient").Return(mockLoggingClient)
+			driver.macAddressMapper.UpdateMappings(test.currentMap)
+
+			if test.name == "duplicate macs" {
```
ajcasagrande: testing via the name is finicky. use the existance of the `test.alternateExpected` field to determine whether or not to enter this code block.

internal/driver/macmapper_test.go
```diff
@@ -178,3 +182,233 @@ func TestMatchEndpointRefAddressToMAC(t *testing.T) {
...
+				ex1 := reflect.DeepEqual(test.expected, driver.macAddressMapper.credsMap)
+				ex2 := reflect.DeepEqual(test.alternateExpected, driver.macAddressMapper.credsMap)
+				assert.Equal(t, true, (ex1 || ex2))
```
ajcasagrande:
```suggestion
				assert.True(t, ex1 || ex2)
```

internal/driver/macmapper_test.go
```diff
@@ -178,3 +182,233 @@ func TestMatchEndpointRefAddressToMAC(t *testing.T) {
...
+		},
+	}
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			m := &MACAddressMapper{
+				credsMu:  sync.RWMutex{},
+				credsMap: test.credMap,
+			}
+			actual := m.ListMACAddresses()
```
ajcasagrande: is this code even used anywhere else? maybe it just needs to be removed.

internal/driver/macmapper_test.go
```diff
@@ -178,3 +182,233 @@ func TestMatchEndpointRefAddressToMAC(t *testing.T) {
...
+		},
+		{
+			mac:      "invalid_mac",
+			expected: "noauth",
+		},
+	}
+
+	driver, mockService := createDriverWithMockService()
+
+	mockService.On("GetLoggingClient").Return(logger.MockLogger{})
```
ajcasagrande:
```suggestion
	mockLogger := logger.NewMockClient()
	mockService.On("GetLoggingClient").Return(mockLogger)
```

internal/driver/macmapper_test.go
```diff
@@ -178,3 +182,233 @@ func TestMatchEndpointRefAddressToMAC(t *testing.T) {
...
+			mac:      "aa:bb:cc:dd:ee:ff",
+			expected: "secret_path",
+		},
+		{
+			mac:      "bb:bb:cc:dd:ee:ff",
+			expected: "default_secret_path",
+		},
+		{
+			mac:      "invalid_mac",
+			expected: "noauth",
```
ajcasagrande:
```suggestion
			expected: noAuthSecretPath,
```

internal/driver/macmapper_test.go
```diff
@@ -178,3 +182,233 @@ func TestMatchEndpointRefAddressToMAC(t *testing.T) {
...
+
+// TestTryGetSecretPathForMACAddress verifies the correct secret path is returned for a given mac address.
+func TestTryGetSecretPathForMACAddress(t *testing.T) {
+
+	tests := []struct {
+		mac      string // input mac address
+		expected string
+	}{
+		{
+			mac:      "aa:bb:cc:dd:ee:ff",
```
ajcasagrande: i think these test cases do need a name its confusing what the different macs mean

internal/driver/macmapper_test.go
```diff
@@ -178,3 +182,233 @@ func TestMatchEndpointRefAddressToMAC(t *testing.T) {
...
+		},
+	}
+
+	driver, mockService := createDriverWithMockService()
+
+	mockService.On("GetLoggingClient").Return(logger.MockLogger{})
+
+	driver.macAddressMapper = NewMACAddressMapper(mockService)
+	driver.macAddressMapper.credsMap = convertMACMappings(t, map[string]string{
+		"secret_path": "aa:bb:cc:dd:ee:ff",
```
ajcasagrande: i would make the `"aa:bb:cc:dd:ee:ff"` as a local variable at the beginning of the func that way it can be referenced so we know it is the mac that has a mapping.

internal/driver/onvifdiscovery_test.go
```diff
@@ -0,0 +1,349 @@
...
+			Name: "testDevice1", Protocols: map[string]models.ProtocolProperties{
+				OnvifProtocol: map[string]string{
+					EndpointRefAddress: "123",
+				},
+			},
+		},
+		{
+			Name: "testDevice2", Protocols: map[string]models.ProtocolProperties{
+				OnvifProtocol: map[string]string{
+					EndpointRefAddress: "456",
```
ajcasagrande: create constants for the endpoint references that are to be used later on.

internal/driver/onvifdiscovery_test.go
```diff
@@ -0,0 +1,349 @@
...
+				},
+			},
+		},
+		{
+			Name: "testDevice3", Protocols: map[string]models.ProtocolProperties{
+				OnvifProtocol: map[string]string{
+					EndpointRefAddress: "789",
```
ajcasagrande: use a uuid generator to use actual uuid values for these fields.

internal/driver/onvifdiscovery_test.go
```diff
@@ -0,0 +1,349 @@
...
+		},
+	}
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			driver, mockService := createDriverWithMockService()
+			mockService.On("Devices").
+				Return(test.devices).Once()
+			mockService.On("Name").
+				Return("device-onvif-camera").Times(test.nameCalls)
```
ajcasagrande: i think the amount of times this is actually called we do not care about. is there a way to ignore that?

internal/driver/onvifdiscovery_test.go
```diff
@@ -0,0 +1,349 @@
...
+// 		Password:           "world",
+// 	}
+// }
+
+// func createDevice() onvif.Device {
+// 	device, _ := onvif.NewDevice(createParams())
+// 	return *device
+// }
+
+// func TestmockService_createDiscoveredDevice(t *testing.T) {
```
ajcasagrande: i think this fails because the code is trying to communicate with an actual camera at that xaddr

ajcasagrande (changes requested):

internal/driver/onvifdiscovery_test.go
ajcasagrande: yes, if you leave off the "Times" part, and do not use any of the `Once`, `Twice`, etc. it will ignore the amount of times it gets called.

internal/driver/driver_test.go
shafinj1: Is this an acceptable comment on function header?  
"// Test to verify the removal of added Onvif device not Onvif control device"

internal/driver/macmapper_test.go
ChristianDarr-personal: You know what, I think you're right. I didn't think much of that. Was it used in an older iteration?

internal/driver/onvifdiscovery_test.go
ChristianDarr-personal: Yes. That is something that I started but will get back to now that I have some more knowledge on mocking

internal/driver/driver_test.go
ajcasagrande: I would write:
```go
// TestDriver_RemoveDevice tests the different code flows of the RemoveDevice when called with an actual device
// versus when called with the control plane device.
```

ChristianDarr-personal (dismissed):
LGTM

internal/driver/credentials_test.go
brianointel: the username and password fields in the test are used for the secret provider mock only; the expected variable is used for the function result comparison.

internal/driver/credentials_test.go
```diff
@@ -7,20 +7,222 @@
...
+				OnvifProtocol: {
+					MACAddress: "",
+				},
+			},
+
+			path:          "default_secret_path",
+			username:      "username",
+			password:      "password",
+			authMode:      onvif.DigestAuth,
+			expected:      Credentials{},
```
presatish: `expected` can be left out here too based on the above comment.

internal/driver/driver_test.go
shafinj1: Comment added

internal/driver/onvifdiscovery_test.go
```diff
@@ -0,0 +1,299 @@
...
+	"testing"
+
+	sdkModel "github.com/edgexfoundry/device-sdk-go/v2/pkg/models"
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/models"
+	contract "github.com/edgexfoundry/go-mod-core-contracts/v2/models"
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	uuid1 = "1793ddc8-28b0-11ed-a261-0242ac120002"
```
presatish, line 19: you can use some uuid generator instead of this?

internal/driver/onvifdiscovery_test.go
```diff
@@ -0,0 +1,299 @@
...
+			Name: "testDevice3", Protocols: map[string]models.ProtocolProperties{
+				OnvifProtocol: map[string]string{
+					EndpointRefAddress: uuid4,
+				},
+			},
+		},
+	}
+}
+
+func TestOnvifDiscovery_makeDeviceMap(t *testing.T) {
```
presatish, line 86: A comment on the test case can be added.

internal/driver/onvifdiscovery_test.go
```diff
@@ -0,0 +1,299 @@
...
+	tests := []struct {
+		name      string
+		devices   []contract.Device
+		deviceMap map[string]contract.Device
+		nameCalls int
```
presatish: What is this for?

internal/driver/onvifdiscovery_test.go
ChristianDarr-personal: I would but Anthony suggested that I make these constants, as they are reused in multiple tests. I don't think I can use uuid.new() to create a constant, it was giving me errors.

internal/driver/macmapper_test.go
```diff
@@ -178,3 +182,159 @@ func TestMatchEndpointRefAddressToMAC(t *testing.T) {
...
+				"creds1": "AA:BB:CC:DD:EE:FF",
+				"creds2": "11:22:33:44:55:66",
+			},
+			expected: map[string]string{
+				"aa:bb:cc:dd:ee:ff": "creds1",
+				"11:22:33:44:55:66": "creds2",
+			},
+		},
+		{
+			name: "single update",
```
presatish: What does `single/no update` mean?

brianointel: @ChristianDarr-personal could you respond to this?

ChristianDarr-personal: I think those are just poorly named tests left over from an earlier version.

ChristianDarr-personal (approved):

internal/driver/macmapper_test.go
presatish: Then it should be removed or renamed.

presatish (approved):
LGTM

ajcasagrande:
This is missing some of the changes to remove the legacy mocking and interfaces. the mock file is gone, but the interface is still there and used.

ajcasagrande (approved):

PR #182, opened 2022-10-23
fix: status delay #121

### LINKED ISSUE #121: Delay in obtaining DeviceStatus for pre-defined and manually added devices
# 🐞 Bug Report

### Affected Services [**REQUIRED**]
device-onvif-camera

### Is this a regression?
No

### Description and Minimal Reproduction [**REQUIRED**]
When a device is added either via pre-defined devices in `camera.toml`, or via the core-metadata REST API, the device will lack a `DeviceStatus` until the next time the `checkStatuses` job is run.

In the case of pre-defined devices (using default configuration) this can be up to 45 seconds after service startup. In the case of REST API, it can be up to 30 seconds.

## 🌍  Your Environment
**Deployment Environment:**
Docker + EdgeX-Compose

**EdgeX Version [**REQUIRED**]:**
2.2

**Anything else relevant?**

ChristianDarr-personal:
This pr closes #121 by adding in a device status check when a device is first added via any method

## PR Checklist

- [x] I am not introducing a breaking change
- [x] I am not introducing a new dependency
- [x] I have added unit tests for the new feature or bug fix
- [x] I have fully tested (add details below) this the new feature or bug fix
- [ ] I have opened a PR for the related docs change - no related docs changes necessary


brianointel:
can you update the PR title to pass the semantic PR check?

internal/driver/checkstatuses.go
```diff
@@ -19,6 +19,11 @@ import (
 // checkStatuses loops through all registered devices and tries to determine the most accurate connection state
 func (d *Driver) checkStatuses() {
 	d.lc.Debug("checkStatuses has been called")
```
brianointel, line 21: is debug too high of a log level for this?

internal/driver/checkstatuses_test.go
```diff
@@ -36,3 +53,156 @@ func TestUpdateDeviceStatus_noUpdate(t *testing.T) {
 	require.NoError(t, err)
 	assert.False(t, changed)
 }
+
```
brianointel, line 56: please remove any leftover commented out code before merging

brianointel (dismissed):
looks good!

internal/driver/checkstatuses.go
ChristianDarr-personal: We felt that sending it to info was too many calls, but we could use a different log level

internal/driver/checkstatuses_test.go
ChristianDarr-personal: Sounds good, I am still playing around with that, but I will remove it tonight if I'm not getting anywhere.

internal/driver/checkstatuses.go
brianointel: i was thinking trace level since it's just indicating the function being called.

ajcasagrande: I think debug is probably fine since it is letting you know the timed trigger occurred.

internal/driver/checkstatuses.go
```diff
@@ -27,40 +32,47 @@ func (d *Driver) checkStatuses() {
...
 	}
-	wg.Wait()
+
+	d.lc.Debugf("device %s status is %s", device.Name, status)
 }
 
 // testConnectionMethods will try to determine the state using different device calls
 // and return the most accurate status
-// Higher degrees of connection are tested first, because if they
-// succeed, the lower levels of connection will too
```
ajcasagrande: should we leave this comment in?

internal/driver/config.go
```diff
@@ -65,7 +66,6 @@ func (c *ServiceConfig) UpdateFromRaw(rawConfig interface{}) bool {
 }
 
 // GetCameraXAddr returns the Address:Port of the camera from the Onvif protocol properties
-// todo: unit test!
```
ajcasagrande, line 68: Is this tested indirectly via your other tests?

ajcasagrande (changes requested):
Overall looks good, a couple questions

internal/driver/checkstatuses.go
presatish: I am ok with debug.

internal/driver/checkstatuses.go
ChristianDarr-personal: I will remove it. I think it's removed on my local machine

internal/driver/config.go
ChristianDarr-personal: I ran into this while stepping through some of the unit tests. It looks like it has a full test dedicated to it, so I just thought I would remove the outdated comment.

ChristianDarr-personal: If it's too unrelated to the current pr, I can just revert it.

internal/driver/checkstatuses.go
```diff
@@ -19,6 +19,11 @@ import (
 // checkStatuses loops through all registered devices and tries to determine the most accurate connection state
 func (d *Driver) checkStatuses() {
 	d.lc.Debug("checkStatuses has been called")
+	start := time.Now()
+	defer func() {
+		d.lc.Debugf("checkStatuses completed in: %v", time.Since(start))
```
presatish, line 24: Not sure if tracking this time is that helpful. I see that this got added because it was removed somewhere else, may be useful for debugging in some cases.

internal/driver/checkstatuses.go
presatish: Should it be removed or still kept in? Seems like it should not be removed.

internal/driver/driver.go
```diff
@@ -457,6 +459,12 @@ func (d *Driver) AddDevice(deviceName string, protocols map[string]models.Protoc
 	if err != nil {
 		return errors.NewCommonEdgeXWrapper(err)
 	}
+
+	device, err := d.sdkService.GetDeviceByName(deviceName)
+	if err != nil {
+		return errors.NewCommonEdgeXWrapper(err)
+	}
+	d.checkStatusOfDevice(device) // check the status of the newly added device
```
presatish: Should the same be done for UpdateDevice() too?

internal/driver/checkstatuses.go
ajcasagrande: I was actually asking if we should _not_ remove it.

internal/driver/config.go
ajcasagrande: It is fine as long as it is outdated. I wanted to make sure the todo was done before deleting it.

internal/driver/driver.go
ajcasagrande: There may be a subset of cases of UpdateDevice where it makes sense to do this, but in general, this will cause an infinite loop as the checkStatus eventually will update the device metadata with status and last seen fields.

internal/driver/checkstatuses.go
ChristianDarr-personal: It was useful when developing, and I think it could also be useful as a debug tool, that time could be used to inform the user the interval at which to run checkStatus for example

internal/driver/checkstatuses.go
ChristianDarr-personal: Right, my bad I had gotten confused on whether I had ultimately removed that line or not.

presatish (dismissed):
LGTM.

ajcasagrande (dismissed):

presatish (dismissed):

lenny-goodell:
@ChristianDarr-personal to fix the DCO issue do the following. This worked for me on another PR
```
git commit -s --amend  (just save the existing commit)
git push --force origin
```

presatish (approved):

PR #301, opened 2023-04-07
feat!: Consume SDK interface changes

weichou1229:
BREAKING CHANGE: Consume SDK interface changes by adding Discover and ValidateDevice func on driver

## PR Checklist

- [ ] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [x] I am not introducing a new dependency (add notes below if you are)
- [ ] I have added unit tests for the new feature or bug fix (if not, why?) not impact
- [x] I have fully tested (add details below) this the new feature or bug fix (if not, why?)
- [ ] I have opened a PR for the related docs change (if not, why?) not impact
  <link to docs PR>

## Testing Instructions

Run core service and device service to verify the func works.

### DIFF
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -208,7 +208,10 @@ func (d *Driver) debouncedDiscover() {
 			d.debounceTimer = nil
 			d.debounceMu.Unlock()
 
-			d.Discover()
+			err := d.Discover()
+			if err != nil {
+				d.lc.Errorf("failed to run device discovery, %v", err)
+			}
 		}()
 	}
 }
@@ -403,7 +406,7 @@ func (d *Driver) createOnvifClient(deviceName string) error {
 }
 
 // Discover performs a discovery on the network and passes them to EdgeX to get provisioned
-func (d *Driver) Discover() {
+func (d *Driver) Discover() error {
 	d.lc.Info("Discover was called.")
 
 	d.configMu.RLock()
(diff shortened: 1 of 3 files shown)


internal/driver/driver.go
```diff
@@ -412,8 +415,7 @@ func (d *Driver) Discover() {
 	d.configMu.RUnlock()
 
 	if !discoveryMode.IsValid() {
-		d.lc.Errorf("DiscoveryMode is set to an invalid value: %s. Refusing to do discovery.", discoveryMode)
-		return
+		return fmt.Errorf("discoveryMode is set to an invalid value: %s. Refusing to do discovery", discoveryMode)
```
ajcasagrande: `discoveryMode` is based on `d.config.AppCustom.DiscoveryMode`, which I believe is still uppercase in the configuration, so I would leave it uppercase in the error message: `DiscoveryMode`

internal/driver/driver.go
```diff
@@ -208,7 +208,10 @@ func (d *Driver) debouncedDiscover() {
 			d.debounceTimer = nil
 			d.debounceMu.Unlock()
 
-			d.Discover()
+			err := d.Discover()
+			if err != nil {
+				d.lc.Errorf("failed to discover device, %v", err)
```
ajcasagrande: I think something more like: `failed to run device discovery` sounds better.

ajcasagrande (changes requested):

presatish (dismissed):

internal/driver/driver.go
weichou1229: Updated

internal/driver/driver.go
weichou1229: Updated

ajcasagrande (approved):

PR #356, opened 2023-05-10
refactor!: updated Snapshot api to get profile token input.

### LINKED ISSUE #122: `GetSnapshot` command is not consistent with other commands
# 🚀 Feature Request

### Relevant Package [**REQUIRED**]
device-onvif-camera

### Description [**REQUIRED**]
Currently, the `GetSnapshot` command is based off of the old `device-camera-go` code [here](https://github.com/edgexfoundry/device-camera-go/blob/5c4f34d1d59b8e25e1a6316661d463e2495d45fe/internal/driver/onvifclient.go#L119), whch will query the device for `GetProfiles`, select the first one, and then return the snapshot based on that.

All other commands such as `GetStreamUri` require the user to provide the `ProfileToken` via the `jsonObject` parameters, making this command not consistent with them, and not support other Profiles than the first one.

presatish:
BREAKING CHANGE: Making Snapshot command consistent by requiring the request to have profile token to provide snapshot response.

## PR Checklist

- [ ] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [x] I am not introducing a new dependency (add notes below if you are)
- [ ] I have added unit tests for the new feature or bug fix (if not, why?)
- [x] I have fully tested (add details below) this the new feature or bug fix (if not, why?)
- [ ] I have opened a PR for the related docs change (if not, why?)
  <link to docs PR>

## Testing Instructions
make test
make docker
/compose-builder$ git checkout main
/compose-builder$ make pull
/compose-builder$ make gen no-secty ds-onvif-camera
Replace onvif service docker image in docker-compose.yml with your latest image created using make docker from the previous step
/compose-builder$ make up
/bin/configure-subnets.sh without error
/bin/map-credentials.sh without error
Onvif devices should be registered in Edgex
Get the updated postman collection and env from the doc folder, update the env with the names of onvif cameras and execute `Snapshot` command with different profiles or no profiles and test.

### DIFF
--- a/cmd/res/profiles/camera.yaml
+++ b/cmd/res/profiles/camera.yaml
@@ -825,3 +825,10 @@ deviceCommands:
     resourceOperations:
       - { deviceResource: "Profiles" }
       - { deviceResource: "StreamUri" }
+  -
+    name: "SnapshotFunction"
+    readWrite: "R"
+    isHidden: false
+    resourceOperations:
+      - { deviceResource: "Profiles" }
+      - { deviceResource: "Snapshot" }
--- a/cmd/res/profiles/camera.yaml
+++ b/cmd/res/profiles/camera.yaml
@@ -818,17 +818,3 @@ deviceCommands:
     resourceOperations:
       - { deviceResource: "MetadataConfiguration" }
       - { deviceResource: "MetadataConfigurationOptions" }
-  -
-    name: "VideoStream"
-    readWrite: "R"
-    isHidden: false
-    resourceOperations:
-      - { deviceResource: "Profiles" }
-      - { deviceResource: "StreamUri" }
-  -
-    name: "SnapshotFunction"
-    readWrite: "R"
-    isHidden: false
-    resourceOperations:
-      - { deviceResource: "Profiles" }
-      - { deviceResource: "Snapshot" }
--- a/doc/postman/device-onvif-camera.postman_collection.json
+++ b/doc/postman/device-onvif-camera.postman_collection.json
@@ -3794,7 +3794,7 @@
 						"method": "GET",
 						"header": [],
 						"url": {
-							"raw": "http://0.0.0.0:59882/api/v3/device/name/{{EDGEX_DEVICE_NAME}}/Snapshot",
+							"raw": "http://0.0.0.0:59882/api/v3/device/name/{{EDGEX_DEVICE_NAME}}/Snapshot?jsonObject={{EDGEX_MEDIA_PROFILE_BASE64}}",
 							"protocol": "http",
 							"host": [
 								"0",
@@ -3810,6 +3810,12 @@
 								"name",
 								"{{EDGEX_DEVICE_NAME}}",
 								"Snapshot"
+							],
+							"query": [
+								{
+									"key": "jsonObject",
+									"value": "{{EDGEX_MEDIA_PROFILE_BASE64}}"
+								}
 							]
 						},
 						"description": "This request returns a snapshot of the video stream at the time the command is given.\n\nIt is returned in a binary format."
@@ -3821,7 +3827,7 @@
 								"method": "GET",
 								"header": [],
 								"url": {
-									"raw": "http://0.0.0.0:59882/api/v3/device/name/{{EDGEX_DEVICE_NAME}}/Snapshot",
+									"raw": "http://0.0.0.0:59882/api/v3/device/name/{{EDGEX_DEVICE_NAME}}/Snapshot?jsonObject={{EDGEX_MEDIA_PROFILE_BASE64}}",
 									"protocol": "http",
 									"host": [
 										"0",
@@ -3837,6 +3843,12 @@
 										"name",
 										"{{EDGEX_DEVICE_NAME}}",
 										"Snapshot"
+									],
+									"query": [
+										{
+											"key": "jsonObject",
+											"value": "{{EDGEX_MEDIA_PROFILE_BASE64}}"
+										}
 									]
 								}
 							},
--- a/internal/driver/onvifclient.go
+++ b/internal/driver/onvifclient.go
@@ -417,12 +417,9 @@ func (onvifClient *OnvifClient) callSubscribeCameraEventFunction(resourceName, s
 
 // callGetSnapshotFunction returns a snapshot from the camera as a slice of bytes
 // The implementation can refer to https://github.com/edgexfoundry/device-camera-go/blob/5c4f34d1d59b8e25e1a6316661d463e2495d45fe/internal/driver/onvifclient.go#L119
-func (onvifClient *OnvifClient) callGetSnapshotFunction(profileToken []byte) ([]byte, errors.EdgeX) {
-	if len(profileToken) == 0 {
-		return nil, errors.NewCommonEdgeX(errors.KindContractInvalid, "no profile token in request", nil)
-	}
-
-	respContent, edgexErr := onvifClient.callOnvifFunction(onvif.MediaWebService, onvif.GetSnapshotUri, profileToken)
+func (onvifClient *OnvifClient) callGetSnapshotFunction(data []byte) ([]byte, errors.EdgeX) {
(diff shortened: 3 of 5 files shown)


cmd/res/profiles/camera.yaml
```diff
@@ -825,3 +825,10 @@ deviceCommands:
     resourceOperations:
       - { deviceResource: "Profiles" }
       - { deviceResource: "StreamUri" }
+  -
+    name: "SnapshotFunction"
+    readWrite: "R"
+    isHidden: false
+    resourceOperations:
+      - { deviceResource: "Profiles" }
+      - { deviceResource: "Snapshot" }
```
lenny-goodell: This one needs to be `Snapshot` so the `Snapshot` command uses it. Rename the resource `SnapshotFrame` to avoid name confusion.

```suggestion
    name: "Snapshot"
    readWrite: "R"
    isHidden: false
    resourceOperations:
      - { deviceResource: "Profiles" }
      - { deviceResource: "SnapshotFrame" }
```

lenny-goodell (changes requested):

cmd/res/profiles/camera.yaml
presatish: But it seems to be already using it with the existing one only. Need to work with the team if the `deviceResource` name has to be changed as we want to keep them consistent as much as possible.

ChristianDarr-personal (dismissed):
LGTM, validated with simulated camera

cmd/res/profiles/camera.yaml
ajcasagrande: I do not think we need this at all? Does it get the snapshot for all profiles?

ajcasagrande: As in, the `Snapshot` command already works without it.

internal/driver/onvifclient.go
```diff
@@ -418,25 +417,12 @@ func (onvifClient *OnvifClient) callSubscribeCameraEventFunction(resourceName, s
...
-	}
-	if len(profilesResp.Profiles) == 0 {
-		return nil, errors.NewCommonEdgeX(errors.KindServerError, "no onvif profiles found", nil)
-	}
-	requestData, edgexErr := snapshotUriRequestData(profilesResp.Profiles[0].Token)
-	if edgexErr != nil {
-		return nil, errors.NewCommonEdgeXWrapper(edgexErr)
+func (onvifClient *OnvifClient) callGetSnapshotFunction(profileToken []byte) ([]byte, errors.EdgeX) {
+	if len(profileToken) == 0 {
+		return nil, errors.NewCommonEdgeX(errors.KindContractInvalid, "no profile token in request", nil)
```
ajcasagrande: what happens if you leave this code out? Don't most onvif commands let the camera respond an error if the user input is bad?

internal/driver/onvifclient.go
```diff
@@ -418,25 +417,12 @@ func (onvifClient *OnvifClient) callSubscribeCameraEventFunction(resourceName, s
...
-	if !ok {
-		return nil, errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf("invalid GetProfilesResponse of type %T for the camera %s", respContent, onvifClient.DeviceName), nil)
-	}
-	if len(profilesResp.Profiles) == 0 {
-		return nil, errors.NewCommonEdgeX(errors.KindServerError, "no onvif profiles found", nil)
-	}
-	requestData, edgexErr := snapshotUriRequestData(profilesResp.Profiles[0].Token)
-	if edgexErr != nil {
-		return nil, errors.NewCommonEdgeXWrapper(edgexErr)
+func (onvifClient *OnvifClient) callGetSnapshotFunction(profileToken []byte) ([]byte, errors.EdgeX) {
```
ajcasagrande: `profileToken` might be misleading here, as the byte array is a json object of: 

```json
{"ProfileToken":"Profile_1"}
```

So just calling it data or something like that is fine I would think, and if we wanted, we could add a comment that it should contain the profile token

ajcasagrande (changes requested):

internal/driver/onvifclient.go
ajcasagrande: In my case, I tricked the code by sending `e30=` which is base64 for `{}` (empty json). It responds with a valid image. And I am OK with that, if that is how the camera treats the command.

ajcasagrande: By this I mean you should remove the check for data, as it is possible that field is optional (either for some cameras or all cameras)

cmd/res/profiles/camera.yaml
lenny-goodell: > But it seems to be already using it with the existing one only. Need to work with the team if the `deviceResource` name has to be changed as we want to keep them consistent as much as possible.

Your Snap shot command isn't using this command. It would have to use `SnapshotFunction` as the command name. So to @ajcasagrande's point does it need a Profile??

Also you could just name this one `SnapShot` which hides the resource with the same name. I would be more explicate and set IsHiden: true on the resource

ajcasagrande: @lenny-intel Aren't these just macros to get more than one resource from a single command? If so, I'm not sure how this would even work, or how the streamuri command above would work either, unless EdgeX is smart enough to dig through the first response and use it as input to the second call

internal/driver/onvifclient.go
presatish: But that json object contains only profileToken. I can change it tough.

cmd/res/profiles/camera.yaml
lenny-goodell: @ajcasagrande , correct on how they work to combine multiple resources in one command. I also was wondering  how they are working since it just reads both resources. It doesn't pass the Profiles to the other command.

doc/postman/device-onvif-camera.postman_collection.json
```diff
@@ -3794,7 +3794,7 @@
 						"method": "GET",
 						"header": [],
 						"url": {
-							"raw": "http://0.0.0.0:59882/api/v3/device/name/{{EDGEX_DEVICE_NAME}}/Snapshot",
+							"raw": "http://0.0.0.0:59882/api/v3/device/name/{{EDGEX_DEVICE_NAME}}/Snapshot?jsonObject={{EDGEX_MEDIA_PROFILE_BASE64}}",
```
lenny-goodell, line 3797: @presatish , @ajcasagrande , this is why the SnapShot resource is working. I.e. the profile is being passed manually.

That Device Command can be removed since it isn't being used.

cmd/res/profiles/camera.yaml
presatish: Honestly I don't understand this or I didn't have too much time to spend understanding this. I just saw the `videoStream` command and made it like that. If this is wrong my guess is even `VideoStream` needs to be changed. As @ajcasagrande mentioned it might be working without it.

lenny-goodell: @presatish , @ajcasagrande , `videoStream` is not used in the PostMan collection. So never been tested? IMO it should be removed along with `SnapshotFunction`

doc/postman/device-onvif-camera.postman_collection.json
ajcasagrande: Yes that was the point of this PR :)

ajcasagrande (approved):
LGTM

internal/driver/onvifclient.go
ajcasagrande: what i mean is that i correlate a parameter called "profileToken" to be a strong value of the token, and not a json object containing the token as one of the fields.

lenny-goodell (approved):
LGTM

vchitike:
Validation looks good. Tested without any profile token input and it generated snapshot with default/random profile.
Verified that it does not generate snapshot with invalid profile token or invalid json value. Also verified that it generates snapshot for specific valid profile token.

PR #361, opened 2023-05-11
docs: add operationId, and do not remove schemas for Get commands

### LINKED ISSUE #360: Swagger file is missing definitions for some schemas
# 📚 Docs or Wiki Bug Report

### Description [**REQUIRED**]

Mostly related to jsonObject schemas, there is a link to the schema but it has been incorrectly removed from the swagger file.

### What's the affected URL? [**REQUIRED**]
 https://docs.edgexfoundry.org/...

https://docs.edgexfoundry.org/3.0/microservices/device/supported/device-onvif-camera/Walkthrough/general-usage/

One example is `device_GetCapabilities`

## 🔥 Exception or Error
<pre><code>

</code></pre>

ajcasagrande:
Fixes #360 

## PR Checklist

- [X] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [X] I am not introducing a new dependency (add notes below if you are)
- [ ] I have added unit tests for the new feature or bug fix (if not, why?)  **N/A**
- [X] I have fully tested (add details below) this the new feature or bug fix (if not, why?)
- [ ] I have opened a PR for the related docs change (if not, why?) **N/A**
  <link to docs PR>

## Testing Instructions

- Load this version of swager yaml into https://editor.swagger.io/
- verify no schema errors
- verify the schemas exist for GET commands

### DIFF
--- a/doc/openapi/v3/device-onvif-camera.yaml
+++ b/doc/openapi/v3/device-onvif-camera.yaml
@@ -1843,7 +1843,6 @@ paths:
                     event:
                       properties:
                         readings:
-                          uniqueItems: true
                           items:
                             allOf:
                             - $ref: '#/components/schemas/ObjectReading'
--- a/doc/openapi/v3/device-onvif-camera.yaml
+++ b/doc/openapi/v3/device-onvif-camera.yaml
@@ -275,6 +275,7 @@ paths:
               examples:
                 503Example:
                   $ref: '#/components/examples/503Example'
+      operationId: device_GetCapabilities
       externalDocs:
         description: Onvif Specification
         url: https://www.onvif.org/ver10/device/wsdl/devicemgmt.wsdl#op.GetCapabilities
@@ -396,6 +397,7 @@ paths:
         '423': *id003
         '500': *id004
         '503': *id005
+      operationId: device_GetServiceCapabilities
       externalDocs:
         description: Onvif Specification
         url: https://www.onvif.org/ver10/device/wsdl/devicemgmt.wsdl#op.GetServiceCapabilities
@@ -481,6 +483,7 @@ paths:
         '423': *id003
         '500': *id004
         '503': *id005
+      operationId: media_GetServiceCapabilities
       externalDocs:
         description: Onvif Specification
         url: https://www.onvif.org/ver10/media/wsdl/media.wsdl#op.GetServiceCapabilities
@@ -557,6 +560,7 @@ paths:
         '423': *id003
         '500': *id004
         '503': *id005
+      operationId: ptz_GetServiceCapabilities
       externalDocs:
         description: Onvif Specification
         url: https://www.onvif.org/ver20/ptz/wsdl/ptz.wsdl#op.GetServiceCapabilities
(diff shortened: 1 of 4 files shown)


ajcasagrande:
> Currently failing openapi schema validation due to `PullPointSubscription`  and `BaseNotificationSubscription` both using the same internal function name

**UPDATE:** Fixed.

doc/openapi/v3/device-onvif-camera.yaml
```diff
@@ -8242,20 +8371,49 @@ components:
...
         SupportedAnalyticsModules:
           $ref: '#/components/schemas/onvif_SupportedAnalyticsModules'
       required:
       - SupportedAnalyticsModules
       type: object
+    analytics_GetSupportedRules:
+      properties:
+        ConfigurationToken:
+          description: "References an existing Video Analytics configuration. The\
+            \ list of available tokens can be obtained\n\t\t\t\t\t\t\t\t\tvia the\
```
presatish: Looks like there are some extra tabs may be its because its auto generated. The extras should be removed?

doc/openapi/v3/device-onvif-camera.yaml
```diff
@@ -8216,6 +8311,19 @@ components:
       - ConfigurationToken
       - RuleName
       type: object
+    analytics_GetAnalyticsModuleOptions:
+      properties:
+        ConfigurationToken:
+          description: Reference to an existing AnalyticsConfiguration.
```
presatish, line 8356: Is this the right description? I see for `ConfiguationToken` couple of places down where it describes as some `Reference` but in other places it starts of describing as a `token` such as https://github.com/edgexfoundry/device-onvif-camera/pull/361/files#diff-9d76c2f3691e66088d86858515a5e47b0c77fe7b6f7bc178f35430a477ead31cR12287

presatish (changes requested):

ChristianDarr-personal (commented):
Schema is messed up for both get and set network configuration

ajcasagrande:
> Schema is messed up for both get and set network configuration

@ChristianDarr-personal Thanks, I will take a look. That command is an EdgeX macro command, so we may have to add a special case to handle it (or just tweak the sidecar yaml).

doc/openapi/v3/device-onvif-camera.yaml
ajcasagrande: @presatish I would say that yes it is the right description. It is copied directly from the onvif xml schemas (just converted to json-schema/yaml). In my eyes, a `Token` is also a `Reference` and an `ID`. It is merely a way to identify a specific "something" (in this case a AnalyticsConfiguration). I believe onvif schema use these terms interchangeably as well. 

Also, I believe the token you linked is not even the same token, since one is for metadata and the other is analytics, but I could be wrong here.

doc/openapi/v3/device-onvif-camera.yaml
ajcasagrande: I will take a look at if this is a bug in the translation, or if the original had them as well. Either way we can probably replace them, however I would be careful about that, as there may be some places they use the tabs to format for other reasons than block-indenting (which is what they are using it for here).

presatish: In the UI it looks good, the extras seem to be causing no issues.

ajcasagrande: @presatish I have addressed this. They are now multi-line raw strings (and have had the tabs removed)

ChristianDarr-personal (approved):
LGTM

presatish (approved):
LGTM

presatish (approved):
LGTM

PR #475, opened 2024-11-11
fix: Ensures that the parsed value is within the bounds of int32

weichou1229:
Fixes [https://github.com/edgexfoundry/device-onvif-camera/security/code-scanning/2](https://github.com/edgexfoundry/device-onvif-camera/security/code-scanning/2)

To fix the problem, we need to ensure that the integer value parsed from the string is within the bounds of `int32` before converting it. This can be done by using `strconv.ParseInt` with a bit size of 32, which directly parses the string into an `int32` value, or by adding explicit bounds checks before the conversion.

In this case, we will use `strconv.ParseInt` with a bit size of 32 to parse the `MessageLimit` value. This ensures that the parsed value is within the bounds of `int32`.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._


cloudxxx8 (approved):
LGTM

